PR #6, opened 2015-09-10
Logging dissabled

tblah:
New installs will no longer log conversations.
    
Logging seems inappropriate for privacy-oriented software. It is my opinion that logging is not a feature with much benefit to most users but it puts them all at risk. It also seems rude to log the words of whoever you are talking to without first asking their permission.

### DIFF
--- a/util.c
+++ b/util.c
@@ -609,7 +609,7 @@ typedef struct
     uint16_t window_x, window_y, window_width, window_height;
     uint16_t proxy_port;
     uint8_t proxyenable;
-    uint8_t logging_enabled : 1;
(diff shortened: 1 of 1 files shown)


GrayHatter:
@tblah if you can get anyone to agree with you, and then either merge or rebase and I'll merge this.

HandsomeMatt:
logging should be off by default for sure - me and a lot of my friends agree

this pr should be rebased for sure though, how'd op even manage that

tblah:
I fixed it here https://github.com/GrayHatter/utox/pull/8

PR #8, opened 2015-09-25
Dissabled logging by default #2

tblah:
Logging does not seem appropriate for a messaging application intended for the
security minded.

I don't know what I messed up before, sorry.

### DIFF
--- a/util.c
+++ b/util.c
@@ -609,7 +609,7 @@ typedef struct
     uint16_t window_x, window_y, window_width, window_height;
     uint16_t proxy_port;
     uint8_t proxyenable;
-    uint8_t logging_enabled : 1;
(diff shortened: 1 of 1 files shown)


util.c
```diff
@@ -609,7 +609,7 @@ typedef struct
     uint16_t window_x, window_y, window_width, window_height;
     uint16_t proxy_port;
     uint8_t proxyenable;
-    uint8_t logging_enabled : 1;
+    uint8_t logging_enabled : 0;
```
GrayHatter: This is the bit size of the uint8_t, this change make this var 0 bits long...

GrayHatter:
closed as stale

PR #9, opened 2015-09-30
tweaks to de translation

aaannndddyyy:


GrayHatter:
Discribe the changes or Get a +1 and I'll merge it!

aaannndddyyy:
Fixed orthography, missing hyphens, where the English original uses the words contact I use the German equivalent in order to be consistent with the source, propper translation of the word screenshot, instead of "dragging a box around an area of the screen" I mark the area, I close into the tray and not next to it, added a missing comma

PR #12, opened 2015-10-05
fix draw_rect_frame’s parameters

soyersoyer:


GrayHatter:
@stal888 has already fixed this for me. I just haven't push it yet. Thanks for taking the time to make this work! Sorry that I'm behind the times

PR #13, opened 2015-10-05
make audio & video init sequential

soyersoyer:
In OS X 10.9.5 I don't see the audio input devices at every ~second program start.
I think it is related to threading. The serialisation of the audio & video init fixes it.

### DIFF
--- a/tox.c
+++ b/tox.c
@@ -667,8 +667,9 @@ void tox_thread(void *UNUSED(args))
         tox_thread_init = 1;
 
         // Start the treads
-        thread(audio_thread, av);
         thread(video_thread, av);
+        while(!video_thread_init) yieldcpu(1);
+        thread(audio_thread, av);
         thread(toxav_thread, av);
 
         //


GrayHatter:
anything with AV need to be biased off the `new_av_api` branch

GrayHatter:
I can copy this over, or you can if you care about the git credit (or want to make my life easy)

or I can just merge this after new_av_api

soyersoyer:
The latter is ok for me.
But why don't you accept this, and later rebase the new_av_api branch on develop?

GrayHatter:
I disagree with rebasing on a fundamental level, but rebasing doesn't work inside a refactor. Especially if there's a lot of changes, you have to remember almost everything about every commit you're rewriting. And it also breaks bisect.

GrayHatter:
this needs a rebase or merge onto develop. Also can you make sure that you still have this bug too with the new refactor?

PR #32, opened 2015-11-07
fixed not losing focus of edit_msg when clicking the messages panel

Doom032:
This fixes selecting text in the friend message log and pressing ctrl+c to copy it.

This might have the annoying side effect of losing edit dialog focus whenever you send an inline picture or a file to a friend. If this turns out to be too much of a usability regression in practice, I recommend making these buttons explicitly return focus to the edit dialog, like the "send message" button already does(in ui_buttons.h).

### DIFF
--- a/ui.c
+++ b/ui.c
@@ -588,8 +588,8 @@ panel_main = {
             .disabled = 1,
             .drawfunc = draw_group,
             .child = (PANEL*[]) {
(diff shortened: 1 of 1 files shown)


GrayHatter:
@Doom032  submit this to the develop branch please?

Doom032:
@GrayHatter Alright, I'll also change it so the scrollbar is on top.

PR #42, opened 2015-11-12
Enable auto changing bitrate for uTox video

GrayHatter:
Generally untested... It seems to work.

@stal888 @tsudoko


Doom032:
When testing my video with Echobot, I notice these things:
 - Quality is definitely better, so it's probably choosing a higher bitrate

I first thought these bugs were caused by this, but I found they're already present in the devlop branch:
 - Frame corruption becomes a lot worse, especially when sharing a larger part of my screen. If the video keeps changing, it's not a huge problem, but if you make it stand still on an image, it seems to only take a few seconds to become corrupt and unviewable. And it seems to stay this way until there's movement in the frame again, maybe because that forces a redraw. This problem seems to get worse with a higher bitrate
 - For some reason the "incoming call" sound keeps playing forever when calling EchoBot. In the Debian repo version, the "incoming call" sounds isn't heard at all while calling EchoBot.

GrayHatter:
I'm going to merge this into develop so it can get deeper testing and more eyes.

PR #47, opened 2015-11-14
refactor list searching & filtering, and add previous/next tab shortcuts on Linux and Windows

Doom032:
Sorry for the amount of changes. All I wanted to do is add ctrl+[shift+]tab, ctrl+pgup and ctrl+pgdn for going through the list of friends. Then I found it wasn't easy to do with the weird way filtering and searching was implemented(would require more duplicate code as list_selectchat_filtered and list_draw already had). Also, search_offset and search_unset gave me headaches.

So this is the new system:

We still have the item list (`ITEM *item`) containing all items.
We now also have a `shown_list`, which contains indexes which point to the items from the `item` list. This is the list we use to find what items are currently visible in the user's list.
`update_shown_list()` updates our `shown_list` based on filter and search settings. It needs to be called whenever something relevant to those things is changed.
You can also use `find_item_shown_index(ITEM *it)` to get the index in the `shown_list` of given item, which is handy when for example, switching to the next or previous item in the list.

This improves things by keeping the filtering in one place(`update_shown_list()`), which creates the `shown_list` which can be used in other functions, without them having to do any filtering or use the "offsets" or "unsets" of the search. Also, we can implement something like qTox' sorting to keep online friends at the top without altering their original location with this. The disadvantage is needing to remember to call update_shown_list after anything relevant changes, but you'd only typically do that when for example adding a new filter. That doesn't happen much, and it'll be obvious that you need to update the list when working on that.

Using this system I created previous/next tab shortcuts on Linux and Windows.
TODO:
 - someone who knows Objective-C can add the shortcuts for OSX
 - make the list scroll up/down automatically when you switch to a friend not immediately visible

### DIFF
--- a/cocoa/interaction.m
+++ b/cocoa/interaction.m
@@ -285,7 +285,6 @@ static inline void select_right_to_char(char_t c) {
     char n = 0;
     if (theEvent.charactersIgnoringModifiers.length == 1 && (theEvent.modifierFlags & NSDeviceIndependentModifierFlagsMask) == NSControlKeyMask) {
         switch (n = [theEvent.charactersIgnoringModifiers characterAtIndex:0]) {
-            case '0':
             case '1':
             case '2':
             case '3':
@@ -295,7 +294,11 @@ static inline void select_right_to_char(char_t c) {
             case '7':
             case '8':
             case '9':
-                list_selectchat(n - '0');
+                list_selectchat(n - '1');
+                redraw();
+                break;
+            case '0':
+                list_selectchat(9);
                 redraw();
                 break;
             default:
--- a/friend.c
+++ b/friend.c
@@ -177,6 +177,22 @@ void friend_addmessage(FRIEND *f, void *data) {
     }
 }
 
+_Bool friend_set_online(FRIEND *f, _Bool online) {
+    if (f->online == online) {
+        return false;
+    }
+
+    f->online = online;
+    if(!f->online) {
+        friend_set_typing(f, 0);
+    }
+
+    update_shown_list(); // update the contact list
+
+    return true;
+}
+
+
 void friend_set_typing(FRIEND *f, int typing) {
     f->typing = typing;
 }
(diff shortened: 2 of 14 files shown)


stal888:
cocoa uses control-[0-9] to jump to the Nth friend in the list. you might want to implement those too.

GrayHatter:
Maybe take the quassel approch? Ctrl + \d == set && Alt + \d == switch?

But I have a feeling that's much more complicated, so you can make them both just switch to the current number in the view

Doom032:
I added it now as I see it in the obj-c code(alt/ctrl+0 brings you to first chat, alt/ctrl+1 to next, etc.) but I notice it's a bit weird from a user perspective. Because the 1 is the leftmost number on your keyboard, and the zero the rightmost, I think it should be that alt+1 brings you to the first chat, and alt+0 to the 10th. It makes most sense from a user perspective.

Waiting for @stal888 's opinion on this, because that would require changing obj-c code as well(worst would be that it works differently on both OSes).

GrayHatter:
> Waiting for @stal888 's opinion on this, because that would require changing obj-c code as well(worst would be that it works differently on both OSes).

I'm not strictly opposed to that, but ideally it should be in the case of maintaining OS quirks/style

Doom032:
@GrayHatter Obviously, but I don't think this has to do with OS quirks, I don't know OSX that well but I'd assume it'd also go for the user-friendly(with our keyboards) approach of starting with 1 in such cases, and seeing 0 as 10.

stal888:
@Doom032 Yeah, go ahead and do that. If you add the Alt key please check for an active edit first, because OS X binds Alt-[0-9] to insert symbols.

GrayHatter:
@stal888 @Doom032 that reminded me of another uTox issue. Let's completely drop alt + * unless you can tell me it' mandatory. There's a char insertion error using alt on windows right now that I need to fix before we go messing with alt + key,

Doom032:
As I said in IRC, I think I fixed that issue while working on this. Gonna commit this now.

GrayHatter:
fucket, merged now. Looks good to me

PR #52, opened 2015-11-16
do not block video device when video is not used

cebe:
This is a rework of https://github.com/notsecure/uTox/pull/1101 on top of the new AV API.

> initially open a device when it is selected but close it afterwards to
> allow other programs to used it.
> A device is now only opened if it is actually needed, meaning either a
> preview window is opened OR a call is started.

fixes https://github.com/notsecure/uTox/issues/628

### DIFF
--- a/video.c
+++ b/video.c
@@ -3,6 +3,7 @@
 static vpx_image_t input;
 
 static _Bool openvideodevice(void *handle) {
+    // initialize video (will populate video_width and video_height)
     if(!video_init(handle)) {
         debug("video_init() failed\n");
         return 0;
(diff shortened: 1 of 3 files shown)


video.c
```diff
@@ -79,40 +95,63 @@ void video_thread(void *args) {
                         }
```
cebe: @GrayHatter I am unsure about this if statement, imo it should always break here if the NULL device is selected, not only when a param is set.

GrayHatter: I can look deeper soonish, but I agree. You should make it break here, with a comment as to why you changed it incase I missed something.

cebe: fixed by fda1a23

cebe:
@GrayHatter what does "Needs +2" mean?

GrayHatter:
needs 2 other people to test it out. I'll be one as soon as I get a chance to actually look through your changes. Then just need another human to tell me it works too.

GrayHatter:
you can tag anyone who's commented in the issue from notsecure's repo. Or I'll find someone on IRC once I look at it

GrayHatter:
someone else :+1:  this and it's good to go

stal888:
Bug with video preview:

- Open video preview
- Set webcam to none
- Preview window doesn't close.

Then you can cause another bug afterwards:

- Set webcam to something
- Preview again
- Stop preview
- Video device isn't closed.

This was tested on mac; someone should try on windows to make sure the bug is in @cebe's code and not mine.

GrayHatter:
@stal888 One of them I'm not sure which also exists in develop

cebe:
@stal888 good catch, even though the preview window does close in my case the video device seems to be not closed after the second set of steps. Will look into this.

cebe:
fixed. But there is another issue that also exists in develop, created a separate issue for that: #56

stal888:
Good to merge

PR #53, opened 2015-11-16
Add & fix some tooltips, slight cleanup of UI strings

Doom032:
- Fixed search friends tooltip
 - Added tooltips for status, audio call, video call, sending a file, and joining audio chat in a groupchat.
 - Renamed button_chat_left and button_chat_right to button_send_file and button_send_screenshot
 - Cleaned up unused UI strings: CONTACTS_FILTER_ALL, CONTACTS_FILTER_ONLINE, CONTACTS_FILTER_EDIT_HINT, BUTTON_ADD_FRIEND and BUTTON_ACCEPT_FRIEND

Editing all those langs/*.h files was really annoying, and made me realize many of them are in different states of up-to-dateness. It's clear we need some script to deal with it. Would it be good if I made some sort of update_translations.py script using https://github.com/eliben/pycparser ? I could also do without pycparser, but that'd mean either 1) trying to hackishly interpret the C itself(making lots of assumptions about how the file is structured) 2) creating a new file with all the internationalization declarations and their English translations, that could be easily parsed to create the langs/en.h and ui_i18n_decls.h headers. It might take a while either way, but it'd make it easier for translators to contribute. They wouldn't even need to be able to compile µTox.


GrayHatter:
Needs a +1 and I'll merge it, look good to me thanks!

cebe:
:+1:

PR #61, opened 2015-11-21
Move text from the search field to the ID field after clicking "add" instead of syncing it between both fields

tsudoko:
Keeping the text in the search field after clicking "add" makes no sense, in most cases the user would clear the field after adding anyway. I doubt anyone would want to edit the ID from the search field when the focus is on the ID field.


GrayHatter:
@tsudoko please rebase

stal888:
:+1:

PR #66, opened 2015-11-22
updated manpage and added --help option to xlib main.c

cebe:
Could have added it for windows and mac but I have no environment to compile and test those files.

### DIFF
--- a/xlib/main.c
+++ b/xlib/main.c
@@ -1147,6 +1147,15 @@ int main(int argc, char *argv[]) {
                         }
                     }
                 }
+            } else if(strncmp(argv[i], "--help", 6) == 0) {
+                printf("µTox - Lightweight Tox client version %s.\n\n", VERSION);
+                printf("The following options are available:\n\n");
+                printf("  --theme=<theme-name>  Specify a UI theme, where <theme-name> can be one of default, dark, light, highcontrast, zenburn.\n");
+                printf("  --portable            Launch in portable mode: All data will be saved to the tox folder in the current working directory.\n");
+                printf("  --set=<option>        Set an option: start-on-boot, show-window, hide-window.\n");
+                printf("  --version             Print the version and exit.\n");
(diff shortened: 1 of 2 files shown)


xlib/main.c
```diff
@@ -1120,7 +1120,7 @@ int main(int argc, char *argv[]) {
                 return 1;
             }
             if(!strcmp(argv[i], "--version")) {
-                debug("%s\n", VERSION);
+                printf("%s\n", VERSION);
```
GrayHatter: why?

cebe: I thought that debug may be removed or silenced in stable versions or something. This output should be enabled regardless of debug mode or not.

GrayHatter:
@cebe please change it back to debug? I have plans to add additional debug macros for different levels of verbosity

xlib/main.c
GrayHatter: @cebe please change it back to debug? I have plans to add additional debug macros for different levels of verbosity

cebe: will do.

GrayHatter:
:+1: 

@cebe this is probably okay, but using printf instead of debug breaks support for android, and likely for windows as well, as windows will buffer output, and hold it until whenever it wants.

cebe:
as far as I see, windows has its own [main.c](https://github.com/GrayHatter/uTox/blob/develop/windows/main.c) and android too: https://github.com/GrayHatter/uTox/blob/develop/android/main.c

As far as I understood my changes do not have any effect on those platforms but I may be wrong.

GrayHatter:
@cebe no, it won't stop them from compiling, but you can't use printf on android. I.e. your code can't be copied over to android. Not a problem as I don't see anyone needing to run --help in android.

cebe:
I think it if fine this way then.

stal888:
@GrayHatter your call.

PR #69, opened 2015-11-27
Update cn.h

giwhub:


GrayHatter:
@giwhub the `//` at the start of each line means `this is a comment, ignore everything after` so if you want to enable those translations, you need to remove the `//` at the start of each line.

giwhub:
thank you.

langs/cn.h
```diff
@@ -151,13 +150,11 @@ msgstr("代理 (SOCKS 5)")
 msgid(WARNING)
 msgstr("修改网络或代理设置将会断开tox网络连接")
```
stal888: "Tox"?

stal888:
@GrayHatter I can't read Chinese but I went over it with rikaikun and it looks ok to me.

GrayHatter:
thanks @giwhub @stal888

Issue #68, opened 2015-11-27
[HELP NEEDED] fix .travis.yaml to check build of uTox for PRs


GrayHatter:
Travis is partially setup, and just needs a little bit of tweaking so that it can auto build and test PRs

It'd be nice to set it up so it can build Windows, Linux, OSx, and Android.

GrayHatter:
just merged in #70 but I'd still like to use the packages from build.tox.chat too...

PR #71, opened 2015-11-27
use blobs from build.tox.chat

stal888:


GrayHatter:
@stal888 cheers mate

closes #68

Issue #73, opened 2015-11-28
Make file doesn't work on Windows


GrayHatter:
@cebe @Doom032 @stal888  I don't remember if it was either of you, (hopefully it was and not someone else I forgot about) But IIRC, someone offered to get the make file working with Windows? There's also that nasty issue of `make` not working when there's a change to any .h file, requiring `make clean` first. Someone also offered to fix that as well?

Doom032:
I have a Makefile I use personally to cross-compile to Windows. Would probably work if cygwin or mingw on Windows too, but requires you to download the dependencies in the right directories yourself. I made it almost a year ago to test my avatars easily, but figured it was too messy to include into the repo. I still use it though, because it's handier than the build script. Here it is: http://pastebin.com/LJU677vk

Oh yeah, I commented out audio filtering because it didn't compile with it and I didn't know why(bad error message), and because it's my personal file I didn't care about finding out.

I based the Makefile on what Jenkins does.

cebe:
> There's also that nasty issue of make not working when there's a change to any .h file, requiring make clean first.

guess that was me, will look into that. I have no windows around so I can not fix the main issue.

GrayHatter:
fixed in develop

PR #77, opened 2015-12-01
adjust makefile to recompile on header change

cebe:
the header file dependencies can not be easily represented in the
Makefile, but this forces a recompile instead of assuming no change is needed.

see #73


GrayHatter:
@cebe you're awesome mate, Thanks!

cebe:
you're welcome :)

PR #85, opened 2015-12-07
Add encrypt save file support.

GrayHatter:


tox.c
```diff
@@ -306,7 +331,7 @@ static void set_callbacks(Tox *tox)
     utox_set_callbacks_for_transfer(tox);
 }
 
-static size_t load_save(uint8_t **out_data){
+static size_t load_save_data(uint8_t **out_data){
```
cebe: this function could have a better name. current one is confusing. does it load or save data? Maybe `load_savefile_data()`?

ui.c
```diff
@@ -252,6 +252,17 @@ static void draw_add_friend(int UNUSED(x), int UNUSED(y), int UNUSED(w), int hei
     }
 }
 
+/* Draw add a friend window */
```
cebe: comment does not seem to match the function.

tox.c
```diff
@@ -357,71 +381,66 @@ static void tox_after_load(Tox *tox)
...
     memcpy(path_tmp + (path_len - 1), ".tmp", sizeof(".tmp"));
+    debug("Writing tox_save to: '%s'\n", (char*)path_tmp);
 
-    debug("Writing tox_save to: '%s': ", (char*)path_tmp);
-    file = fopen((char*)path_tmp, "wb");
-    if(file) {
-        fwrite(data, size, 1, file);
-        flush_file(file);
-        fclose(file);
+    if (edit_profile_password.length > 3) {
```
cebe: Am I correct that when I have a 3 char or less password, it will just write out the file in clear text?

GrayHatter: That is correct. I'm planning to add text that will prompt the user. But can't decide how I wanted to phrase it, but wanted to get some eyes on the core of the code now.

tsudoko:
Some ideas:

 - password field should obscure the text (****, ●●●●, █ █ █ █  or similar)
 - password should not be easily changeable from the settings (make the field disabled, require clicking a button to enable it and clicking a button to change the password, perhaps require typing the password twice too)
 - don't try to decrypt the save as the user is typing - require clicking a button, a delay after some failed attempts would be good as well

Some mockups:
image
image
image

stal888:
Maybe all the password stuff should go in its own tab

GrayHatter:
I was talking with rowan_stripe in IRC, and he said he was a little weird that the password auto submits. But other than it's not standard, I can't think of a reason to add buttons at all. I think swapping the characters to dots, or [other], on lose focus is a good idea, but I can't think of a reason other than it's the standard to have to press enter.

I'm also not going to implement, a delay after failed attempts. The goal of that is to prevent brute forcing, but you can't use uTox to brute force faster at most you'd be able to try 60/m. I'm going to need convincing that it needs to be less than that.

- [X] Profile encryption
- [ ] chat log encryption
- [ ] Warn the user about unencrypted profiles.
- [ ] Obscure text without focus
- [ ] add a button to disconnect tox, and go back to the password screen
- [ ] add a differnt user badge/avatar/name/box when not connected

LittleVulpix:
Most people I know that use this type of software strongly dislike no confirmation of change. A button that becomes unclickable after being clicked is perhaps seemed as unnecessary, but it fills the user with confidence. When there is no button (such as 'OK' or 'APPLY'), there must be very clear indication that the changes have been made.

It is probably slightly annoying for normal settings, but password is a special (more important) case whereby the user risks losing (being unable to decrypt/access) all of his/her data if the password is not saved properly. Please reconsider adding a button - user will always look for one.

GrayHatter:
The more I think about it the more I agree. And @tsudoko has all but
convinced me in IRC last night too.

GrayHatter:
@tsudoko @stal888 @LittleVulpix this is ready for develop, and in needed for refactoring message.c

tsudoko:
When the password field contains <3 characters: `Using encrypted data, trying password: Unknown error, please file a bug report!`

GrayHatter:
@tsudoko good point, thanks

tsudoko:
You can change your avatar when the profile is locked.

util.c
```diff
@@ -49,7 +49,10 @@ void file_write_raw(uint8_t *path, uint8_t *data, size_t size)
         return;
     }
 
-    fwrite(data, size, 1, file);
+    size_t outsize;
+    outsize = fwrite(data, sizeof(uint8_t), size, file);
```
cebe: unused variable `outsize`

ui.c
```diff
@@ -57,40 +57,50 @@ void draw_avatar_image(UTOX_NATIVE_IMAGE *image, int x, int y, uint32_t width, u
...
-
-    /* Draw online/all friends filter text. */
-    setcolor(!button_filter_friends.mouseover ? COLOR_MENU_SUBTEXT : COLOR_MAIN_HINTTEXT);
-    setfont(FONT_STATUS);
-    drawtextrange(SIDEBAR_FILTER_FRIENDS_LEFT, SIDEBAR_FILTER_FRIENDS_WIDTH, SIDEBAR_FILTER_FRIENDS_TOP,
-                  list_get_filter() ? S(FILTER_ALL)    : S(FILTER_ONLINE),
-                  list_get_filter() ? SLEN(FILTER_ALL) : SLEN(FILTER_ONLINE) );
+        setcolor(!button_name.mouseover ? COLOR_MENU_TEXT : COLOR_MENU_SUBTEXT);
+        setfont(FONT_SELF_NAME);
+        drawtextrange(SIDEBAR_NAME_LEFT, SIDEBAR_NAME_WIDTH, SIDEBAR_NAME_TOP, "Not connected!", 14);
```
cebe:
```
  CC    ui.o
ui.c: In function ‘draw_user_badge’:
ui.c:102:80: warning: pointer targets in passing argument 4 of ‘drawtextrange’ differ in signedness [-Wpointer-sign]
         drawtextrange(SIDEBAR_NAME_LEFT, SIDEBAR_NAME_WIDTH, SIDEBAR_NAME_TOP, "Not connected!", 14);
                                                                                ^
In file included from ui.c:1:0:
main.h:356:6: note: expected ‘char_t *’ but argument is of type ‘char *’
 void drawtextrange(int x, int x2, int y, char_t *str, STRING_IDX length);
      ^
```

cebe:
when I unlock my profile, I end up on settings page with focus on the password field and the password is shown in clear text. This should not happen. Also in general I do not think showing the password in clear text while unlocking the profile is a good idea. I may want to unlock it while someone is looking at my screen.

GrayHatter:
just merged develop into this branch so we can merge it for 0.4.6

cebe:
Still getting this:

```
Using encrypted data, trying password: Unknown error, please file a bug report!
Using encrypted data, trying password: Unknown error, please file a bug report!
Using encrypted data, trying password: Unknown error, please file a bug report!
Using encrypted data, trying password: Unknown error, please file a bug report!
Using encrypted data, trying password: Unknown error, please file a bug report!
Using encrypted data, trying password: Unknown error, please file a bug report!
Using encrypted data, trying password: Unknown error, please file a bug report!
Using encrypted data, trying password: Unknown error, please file a bug report!
Using encrypted data, trying password: Unknown error, please file a bug report!
Using encrypted data, trying password: Unknown error, please file a bug report!
Using encrypted data, trying password: Unknown error, please file a bug report!
Using encrypted data, trying password: Unknown error, please file a bug report!
Using encrypted data, trying password: Unknown error, please file a bug report!
...
```

cebe:
Avatar click still opens the file chooser when profile is locked. and it changes the avatar.

cebe:
If I unlock my profile, the password field should not have focus after successful unlock. Otherwise hitting any key accidentally will change the password my profile is encrypted with.

cebe:
on start on a locked profile it would be nice to have password field focused though.

cebe:
UI gets broken if I change network settings:

1. open utox with an encrypted profile
2. unlock
3. go to network settings, switch IPv6 on or off

bildschirmfoto von 2015-12-20 05 54 36

cebe:
Seems unrelated but that bug does not exist in develop: I get a segfault when pasting something into the add friend toxId field on the "Add New Contact" page:

```
SelectionNotify
Type: ATOM
Property: ATOM
Supported type: STRING
Supported type: UTF8_STRING
SelectionNotify
Type: UTF8_STRING
Property: TARGETS

Program received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x0000000000417a8d in edit_paste (
    data=0xe70dd0 "B79C465783226865C40174B1C5B19FCE2EF290AE92FE50A3C20250506A28D1520EE385FE2E77", 
    length=76, select=false) at edit.c:787
#2  0x000000000041cfe2 in pastedata (data=0xe70dd0, type=292, len=76, select=false) at xlib/main.c:835
#3  0x000000000041ef59 in doevent (event=...) at xlib/event.c:475
#4  0x0000000000420609 in main (argc=2, argv=0x7fffffffe288) at xlib/main.c:1447
```

PR #100, opened 2015-12-19
Use "Leave" instead of "Remove Group"

tsudoko:
Fixes notsecure/uTox#1105. This string had been used before and it's translated into many languages already, it would be a waste to use a different one.


cebe:
there is also LEAVE_GROUP, which may fit even better. https://github.com/GrayHatter/uTox/blob/develop/langs/en.h#L280-L281

PR #107, opened 2015-12-22
Fix pointless rebuilding of uTox

albel727:
The makefile uses tray-icon target instead of `$(TRAY_OBJ)`. It causes it to always rebuild, so if you do `make FILTER_AUDIO=1` + `make install` (i.e. the latter without FILTER_AUDIO as the ebuild does), make install fails because it tries to relink utox without libfilteraudio (using objects that were compiled with FILTER_AUDIO=1). To fix it's sufficient to replace both usages of `tray-icon` with `$(TRAY_OBJ)`, as does this commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ GIT_V = $(shell git describe --abbrev=8 --dirty --always --tags)
 
 all: utox
 
-utox: $(OBJ) $(OS_OBJ) tray-icon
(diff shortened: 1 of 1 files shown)


GrayHatter:
this is already fixed in develop, if you really need this pulled into master, use the makefile from develop

PR #123, opened 2015-12-26
Fix compilation on arm

doughdemon:


GrayHatter:
I was told this works... I hope he's right

PR #144, opened 2015-12-31
Centralize argument parsing

tgharib:
Solves https://github.com/GrayHatter/uTox/issues/142 and https://github.com/notsecure/uTox/issues/1060

### DIFF
--- a/src/android/main.c
+++ b/src/android/main.c
@@ -520,20 +520,20 @@ void config_osdefaults(UTOX_SAVE *r)
 
 static void android_main(struct android_app* state){
     bool theme_was_set_on_argv;
-    int8_t launch_at_startup;
+    int8_t should_launch_at_startup;
     int8_t set_show_window;
     bool no_updater;
-    
-    parseArgs(NULL, NULL, &theme_was_set_on_argv, &launch_at_startup, &set_show_window, &no_updater);
-    
-    if (launch_at_startup == 1 || launch_at_startup == -1) {
+
+    parse_args(NULL, NULL, &theme_was_set_on_argv, &should_launch_at_startup, &set_show_window, &no_updater);
+
+    if (should_launch_at_startup == 1 || should_launch_at_startup == -1) {
         debug("Start on boot not supported on this OS!\n");
     }
-    
+
     if (set_show_window == 1 || set_show_window == -1) {
         debug("Showing/hiding windows not supported on this OS!\n");
     }
-    
+
     if (no_updater == true) {
         debug("Disabling the updater is not supported on this OS.\n");
     }
--- a/src/android/main.c
+++ b/src/android/main.c
@@ -519,10 +519,10 @@ void config_osdefaults(UTOX_SAVE *r)
 }
 
 static void android_main(struct android_app* state){
-    _Bool theme_was_set_on_argv;
+    bool theme_was_set_on_argv;
     int8_t launch_at_startup;
     int8_t set_show_window;
-    _Bool no_updater;
+    bool no_updater;
     
     parseArgs(NULL, NULL, &theme_was_set_on_argv, &launch_at_startup, &set_show_window, &no_updater);
     
@@ -534,8 +534,8 @@ static void android_main(struct android_app* state){
         debug("Showing/hiding windows not supported on this OS!\n");
     }
     
-    if (no_updater == 1) {
-        debug("Disabling the updater is not supported on this OS. Updates are managed by the play store.\n");
+    if (no_updater == true) {
+        debug("Disabling the updater is not supported on this OS.\n");
     }
 
     // Make sure glue isn't stripped
--- a/src/android/main.c
+++ b/src/android/main.c
@@ -519,6 +519,24 @@ void config_osdefaults(UTOX_SAVE *r)
 }
 
 static void android_main(struct android_app* state){
+    _Bool theme_was_set_on_argv;
+    int32_t launch_at_startup;
+    int32_t set_show_window;
+    _Bool no_updater;
+    
+    parseArgs(NULL, NULL, &theme_was_set_on_argv, &launch_at_startup, &set_show_window, &no_updater);
+    
+    if (launch_at_startup == 1 || launch_at_startup == -1) {
+        debug("Start on boot not supported on this OS!\n");
+    }
+    
+    if (set_show_window == 1 || set_show_window == -1) {
+        debug("Showing/hiding windows not supported on this OS!\n");
+    }
+    
+    if (no_updater == 1) {
+        debug("Disabling the updater is not supported on this OS. Updates are managed by the play store.\n");
+    }
 
     // Make sure glue isn't stripped
 
--- a/src/android/main.c
+++ b/src/android/main.c
@@ -520,8 +520,8 @@ void config_osdefaults(UTOX_SAVE *r)
 
 static void android_main(struct android_app* state){
     _Bool theme_was_set_on_argv;
-    int32_t launch_at_startup;
-    int32_t set_show_window;
+    int8_t launch_at_startup;
+    int8_t set_show_window;
     _Bool no_updater;
     
     parseArgs(NULL, NULL, &theme_was_set_on_argv, &launch_at_startup, &set_show_window, &no_updater);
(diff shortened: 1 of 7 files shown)


src/android/main.c
```diff
@@ -519,6 +519,24 @@ void config_osdefaults(UTOX_SAVE *r)
...
+    if (launch_at_startup == 1 || launch_at_startup == -1) {
+        debug("Start on boot not supported on this OS!\n");
+    }
+    
+    if (set_show_window == 1 || set_show_window == -1) {
+        debug("Showing/hiding windows not supported on this OS!\n");
+    }
+    
+    if (no_updater == 1) {
+        debug("Disabling the updater is not supported on this OS. Updates are managed by the play store.\n");
```
cebe: not necessarily Play store but "App Store" in general. e.g. fdroid.

src/main.c
```diff
@@ -1,4 +1,104 @@
...
+                } else if (!strcmp(optarg, "dark")) {
+                    theme = THEME_DARK;
+                } else if (!strcmp(optarg, "light")) {
+                    theme = THEME_LIGHT;
+                } else if (!strcmp(optarg, "highcontrast")) {
+                    theme = THEME_HIGHCONTRAST;
+                } else if (!strcmp(optarg, "zenburn")) {
+                    theme = THEME_ZENBURN;
+                } else {
+                    debug("Please specify correct theme (please check user manual for list of correct values).");
```
cebe: could just list valid values here as we have a complete list above anyway.

src/main.c
```diff
@@ -1,4 +1,104 @@
...
+    
+    static struct option long_options[] = {
+        {"theme",         required_argument,   NULL,  't'},
+        {"portable",      no_argument,            NULL,  'p'},
+        {"set",              required_argument,   NULL,  's'},
+        {"unset",          required_argument,   NULL,  'u'},
+        {"no-updater", no_argument,            NULL,  'n'},
+        {"version",       no_argument,            NULL,  'v'},
+        {"help",            no_argument,            NULL,  'h'},
+        {0, 0, 0, 0}
```
benwaffle: AHHHHHH they don't line up

src/main.c
```diff
@@ -1,4 +1,104 @@
...
+                break;
+                
+            case 'u':
+                if (!strcmp(optarg, "start-on-boot")) {
+                    *launch_at_startup = -1;
+                } else {
+                    debug("Please specify a correct unset option (please check user manual for list of correct values).");
+                    exit(EXIT_FAILURE);
+                }
+                break;
```
benwaffle: I think we should remove set & unset. 1) they only handle 3 (or 1) flags and 2) i doubt anybody ever used them
If we were to keep it, it should be able to set & unset any config value

src/main.c
```diff
@@ -1,4 +1,104 @@
...
+            case 'h':
+                printf("µTox - Lightweight Tox client version %s.\n\n", VERSION);
+                printf("The following options are available:\n\n");
+                printf("  -t --theme=<theme-name>  Specify a UI theme, where <theme-name> can be one of default, dark, light, highcontrast, zenburn.\n");
+                printf("  -p --portable            Launch in portable mode: All data will be saved to the tox folder in the current working directory.\n");
+                printf("  -s --set=<option>        Set an option: start-on-boot, show-window, hide-window.\n");
+                printf("  -u --unset=<option>      Unset an option: start-on-boot.\n");
+                printf("  -n --no-updater          Disable the updater.\n");
+                printf("  -v --version             Print the version and exit.\n");
+                printf("  -h --help                Shows this help text.\n\n");
```
benwaffle: `puts()` is a tiny bit cleaner

cebe:
travis fails on Mac, btw: https://travis-ci.org/GrayHatter/uTox/jobs/99537541#L2421

src/main.c
tgharib: @cebe Not sure what you mean here.

src/main.c
tgharib: @benwaffle I swear they lined up in NetBeans! NetBeans lied to me. :(

src/main.c
cebe: instead of "please check user manual for list of correct values", say: "valid values are ..."

src/main.c
tgharib: I agree. Do you think we should just have four flags: `enable-start-on-boot`, `disable-start-on-boot`, `show-window` and `hide-window`?

stal888:
Is there a reason you don't use bools for launch_at_startup/set_show_window?

src/main.c
GrayHatter: I'm undecided, I like @cebe's suggestion, but that's just one more place we have to do upkeep...

Also, the manual also tells you how to make a custom one, I'd kinda like to push people that way

GrayHatter:
@stal888 because it's a set, unset. We don't want to change them without that flag so we can't just if() it, else we'll unset when it's null. (I already asked that question)

src/main.c
benwaffle: i don't think any of those options are useful

GrayHatter: Says the Linux user... Those are windows settings.

GrayHatter: Also, if they tray doesn't draw, and uTox is set to start in tray... You're kinda fucked

benwaffle: those settings can be set from the GUI
and if they tray doesn't draw you have a bug to fix

tsudoko:
Why are you using camel case in function names? Other parts of codebase use snake case.

tgharib:
Alright, I've implemented your suggestions and the pull request should be complete. It needs to be tested. I can't seem to compile for windows so I've only tested on linux.

GrayHatter:
@tgharib 
```
  CC    src/windows/main.o
src/windows/main.c: In function ‘CommandLineToArgvA’:
src/windows/main.c:1193:5: warning: suggest parentheses around assignment used as truth value [-Wparentheses]
     while (a = CmdLine[i]) {
     ^
src/windows/main.c: In function ‘WinMain’:
src/windows/main.c:1346:15: error: ‘arglist’ undeclared (first use in this function)
     LocalFree(arglist);
               ^
src/windows/main.c:1346:15: note: each undeclared identifier is reported only once for each function it appears in
src/windows/main.c:1426:5: warning: implicit declaration of function ‘dnd_init’ [-Wimplicit-function-declaration]
     dnd_init(hwnd);
     ^
Makefile:134: recipe for target 'src/windows/main.o' failed
make: *** [src/windows/main.o] Error 1
```

tgharib:
Ping. Hopefully no more compilation errors.

GrayHatter:
testing this today

src/main.c
```diff
@@ -1,4 +1,104 @@
...
+                    debug("Please specify a correct set option (please check user manual for list of correct values).");
+                    exit(EXIT_FAILURE);
+                }
+                break;
+                
+            case 'u':
+                if (!strcmp(optarg, "start-on-boot")) {
+                    *should_launch_at_startup = -1;
+                } else {
+                    debug("Please specify a correct unset option (please check user manual for list of correct values).");
```
tsudoko: needs `\n`

src/main.c
```diff
@@ -1,4 +1,104 @@
...
+                
+            case 's':
+                if (!strcmp(optarg, "start-on-boot")) {
+                    *should_launch_at_startup = 1;
+                } else if (!strcmp(optarg, "show-window")) {
+                    *set_show_window = 1;
+                } else if (!strcmp(optarg, "hide-window")) {
+                    *set_show_window = -1;
+                } else {
+                    debug("Please specify a correct set option (please check user manual for list of correct values).");
```
tsudoko: needs `\n`

src/main.c
tsudoko: needs `\n`

GrayHatter:
themes now work on windows in release-v0.6

tgharib:
Any reason why this has been closed? I couldn't test on Windows due to outdated documentation at the time so if that's the issue, I can take another crack at compiling on Windows.

GrayHatter:
Closed because it's already staged in the branch release-v0.6 and will get
pushed to master tonight

PR #146, opened 2015-12-31
Addition to PR#112

linuxraspilxc:
*fixed heading
+recommendation for an IDE


GrayHatter:
c.learncodethehardway.org/book/ex0.html

GrayHatter:
http://c.learncodethehardway.org/book/ex0.html

PR #150, opened 2016-01-04
Fix OpenBSD compatibility

czarkoff:
In `Makefile`:

* Apply all CFLAGS and LDFLAGS configuration options unconditionally.  Platform-specific overrides may be applied on top of that.
* Don't add `-lresolv -ldl` to OpenBSD, where they are not needed.

In `src/xlib/main.c`:

* Remove Linux-specific mechanism for handling key input.  Xlib-based mechanism is portable and good enough for Linux as well.

I believe that these changes allow building uTox on other BSDs as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,59 +5,48 @@ V4LCONVERT = 1
 FILTER_AUDIO = 0
 UNITY = 0
 
+PKG_CONFIG ?= pkg-config
+
+OUT_FILE = utox
+
 DEPS = libtoxav libtoxcore openal vpx libsodium
 
-UNAME_S := $(shell uname -s)
-UNAME_O := $(shell uname -o)
+UNAME_S != uname -s 2>/dev/null
+UNAME_O != uname -o 2>/dev/null
 
-CFLAGS += -g -Wall -Wshadow -pthread -std=gnu99
+CFLAGS ?= -g -Wall -Wshadow
+CFLAGS += -pthread -std=gnu99
 LDFLAGS += -pthread -lm
 
-ifeq ($(FILTER_AUDIO), 1)
-	DEPS += filteraudio
-	CFLAGS += -DAUDIO_FILTERING
-endif
+DEPS += fontconfig freetype2 x11 xext xrender
+
+OS_SRC = $(wildcard src/xlib/*.c)
+OS_OBJ = $(OS_SRC:.c=.o)
+
+TRAY_OBJ = icons/utox-128x128.o
+TRAY_GEN = $(LD) -r -b binary icons/utox-128x128.png -o
 
 ifeq ($(UNAME_S), Linux)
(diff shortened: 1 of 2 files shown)


GrayHatter:
xlib/main.c is done that way because it's much faster, more reliable, and less prone to interference when playing a game. You'll have to offer something better rather than just stripping it out.

These make file changes break make in windows + cygwin

czarkoff:
I would argue that stripping platform-specific stuff is better then having it, particularly in the project that pretends to be portable. Anyway, I'd look into it. But it would help if you could specify the issues with xlib code in more detail, eg. what causes interference, or which reliability issues you encountered using it.

Re changes in makefile: what is the output of `uname -s` on Cygwin?

GrayHatter:
I have no problem if you want to wrap it in an ifdif, but because you can't use it isn't a reason to just delete it.

uname -s `CYGWIN_NT-10.0`

But the issue is the dependencies that you've added. by removing the linux section

czarkoff:
Does cygwin set `PKG_CONFIG` environment variable?

GrayHatter:
It sets $PKG_CONFIG_PATH

czarkoff:
FWIW I believe this code should build on Cygwin, Linux, OSX and BSDs now.

czarkoff:
Ping. I believe this can be merged now.

GrayHatter:
Other than fixing OpenBSD, why is restructuring Makefile better?

GrayHatter:
Also no, this still doesn't work on cygwin. It can't find linux headers it doesn't need.

This seems to be another reason it's better to make decisions on how to compile by what platform you're running on.

Most of the time I'd agree that when you can strip platform choices, you should. But the Makefile isn't one of them. Every platform compiles in different ways, and they all conflict.

czarkoff:
Basically the idea is that most platform need the same set of rules, and cygwin, as a special case, needs another set of rules. So making only Cygwin stand out will help if, say, someone wants to build uTox on AIX, Haiku, Minix, Solaris, QNX or some other rare platform.

GrayHatter:
Cygwin isn't the only one as a special case, cocoa on OSX is another, the difference being that it's rules are so incompatible it needs it's own makefile.

czarkoff:
I'd rather go for separate Makefile for Cygwin as well.

GrayHatter:
I do most of the development on windows and/or linux, so I'd rather go for a separate Makefile for bsd

czarkoff:
OK, here is separate makefile for BSDs approach.

* Add BSD makefile under name `makefile`.
* Rename `Makefile` to `GNUmakefile` so that GNU make uses it instead of BSD makefile
* Enclose Linux-specific in `src/xlib/main.c` inside `#ifdef LINUX_IO`.

PR #154, opened 2016-01-08
Fix OpenBSD compatibility

czarkoff:
As discussed in #150, here is alternative approach:

 * Add BSD makefile under name `makefile`.
 * Rename `Makefile` to `GNUmakefile` so that GNU make uses it instead of BSD makefile
 * Enclose Linux-specific in `src/xlib/main.c` inside ifdef `LINUX_IO`.


czarkoff:
Sorry, these patches need refactoring. Closing for now.

PR #156, opened 2016-01-09
Fix OpenBSD compatibility

czarkoff:
I enclosed Linux-specific code in `src/xlib/main.c` inside ifdef `LINUX_IO`.
Next, I added BSD makefile under name `makefile`. BSD make looks for `makefile` first, so this name ensures that BSD make will not stumble on GNU makefile and die. GNU make looks for `GNUmakefile`, `makefile` and `Makefile`, in this order. To help it pick right makefile, I renamed `Makefile` to `GNUmakefile`.

BSD makefile allows several parallel builds are possible off the same repo, in different directories. It also provides more control over installation paths. It should be usable on Linux with `bmake`, on BSDs and (possibly) on Minix.

(Sorry for repeated pull requests - forced pushing didn't cause such problems for me before.)

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -19,6 +19,8 @@ ifeq ($(FILTER_AUDIO), 1)
 endif
 
 ifeq ($(UNAME_S), Linux)
+	CFLAGS += -DLINUX_IO
+
 	OUT_FILE = utox
 
 	DEPS += fontconfig freetype2 x11 xext xrender
--- a/docs/BUILD.md
+++ b/docs/BUILD.md
@@ -69,6 +69,40 @@ sudo ldconfig
 Have fun!
 
 If you're looking for a good IDE, Netbeans is very easy to setup for uTox, in fact, you can just create a new project from the existing sources and everything should work fine.
+
+
+### BSD
+
+uTox distribution includes BSD makefile, which works differently from GNU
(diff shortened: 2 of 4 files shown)


czarkoff:
@GrayHatter ping.

GrayHatter:
Just busy, haven't forgotten or decided.

cebe:
is it not possible to have one makefile for all systems?

cebe:
> is it not possible to have one makefile for all systems?

okay, just seen https://github.com/GrayHatter/uTox/pull/150#issuecomment-169870221

GrayHatter:
https://www.gnu.org/software/make/manual/make.html#Recursion

@czarkoff What about having a single makefile that when invoked without a target tries to guess what platform it's on, and pulls that makefile with `make -C ./somedir/Makefile.{Windows,Linux,OSX,BSD}` or when invoked with a target such as `make BSD` ignores the platform and pulls that makefile?

czarkoff:
I like this idea. I'll open another PR with implementation draft.

GrayHatter:
maybe put all the makefiles into `./tools/`

I was thinking that they should go into their src/ directories, but I'd rather them all be in the same place so that if you're going to edit one, you can edit them all.

Issue #157, opened 2016-01-09
AltGr+number must not switch tab


rvalles:
AltGr+number row is used in many keymaps (including the spanish keymap) to type relatively common symbols such as @,~,#,€,¬.

Therefore, AltGr+number must not switch tab.

GrayHatter:
whoops... @stal @Doom032

GrayHatter:
#47

stal888:
cocoa uses the ctrl key, so we're good there.

I think there's only one AltGr key on normal keyboards, so you could exclude that one from the mask @GrayHatter

Doom032:
@rvalles You're on Windows? Because I don't have this problem on Linux.

If so, @GrayHatter looking at the Windows documentation, the fix is probably to replace VK_MENU with VK_LMENU in windows/main.c:1612. I had expected VK_MENU to only match true on the actual Alt key, not Altgr, but I guess I was mistaken. I can't test on Windows though, so you may put in the fix.

rvalles:
@Doom032: Issue doesn't happen with Linux indeed. Only Windows.

rvalles:
This (altgr+number on Windows) is still an issue as of 0.6.1, which I've just upgraded to.

GrayHatter:
can't reproduce this...

Are you sure you're on v0.6.1?

Where'd you get it?

Is it still affected?

rvalles:
Yes, I can reproduce it. AltGr + Number row changes tabs.

* Windows 10.
* Happens with Catalan, Spanish, Irish input languages. Doesn't happen with Japanese (which is a bit special to begin with).
* 0.6.1 was obtained via the bundled updater.

rvalles:
Furthermore (despite 0.6.1 in title bar), I've just done this test:

* Deleted utox (Roaming/uTox)
* Redownloaded and ran utox runner.

Issue persists.

GrayHatter:
can you compile utox?

rvalles:
Got VS2015, but haven't messed with mingw (which build seems to require) for a very long time; I'd have to look into it.

Do you think the version available via utox_runner wasn't really built from 0.6.1?

Let me know if you _really_ want me to try and build it.

GrayHatter:
I was thinking about adding some more debug info so that I could see what
button is really getting pressed. Because that really shouldn't be happening

GrayHatter:
can you check what key is being pressed http://www.keyboardtester.com/

[...]

PR #160, opened 2016-01-13
User now can hide app before making screenshots, fixed bug with 'send…

ingvar1995:
User now can hide app before making screenshots, fixed bug with 'sending' inline to offline contact, alt + num disabled in windows ( #157 ), updated Russian translation.


GrayHatter:
this is awesome, I'll use them all, but they need to be separate commits and Pull requests

ingvar1995:
ok. i need to check my code first

PR #166, opened 2016-01-14
Fixed some bugs

ingvar1995:
Fixed bug with contacts, alt + num disabled in win ( #157 ), fixed bug with sending inline to offline contact

### DIFF
--- a/src/roster.c
+++ b/src/roster.c
@@ -771,7 +771,7 @@ _Bool list_mup(void *UNUSED(n)) {
     _Bool draw = 0;
     tooltip_mup(); /* may need to return one true */
     if(selected_item_mousedown && abs(selected_item_dy) >= 5) {
-        if(nitem) {
+        if(nitem && find_item_shown_index(nitem) != INT_MAX) { 
             if(selected_item->item == ITEM_FRIEND) {
                 if(nitem->item == ITEM_FRIEND) {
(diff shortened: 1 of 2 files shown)


src/windows/main.c
```diff
@@ -730,6 +730,10 @@ void paste(void)
     if(!h) {
         h = GetClipboardData(CF_BITMAP);
         if(h && selected_item->item == ITEM_FRIEND) {
+			FRIEND *f = selected_item->data;
```
stal888: Indentation should be 4 spaces each here

ingvar1995:
smth strange happened with that line. looks like a bug

src/roster.c
```diff
@@ -771,7 +771,7 @@ _Bool list_mup(void *UNUSED(n)) {
     _Bool draw = 0;
     tooltip_mup(); /* may need to return one true */
     if(selected_item_mousedown && abs(selected_item_dy) >= 5) {
-        if(nitem) {
+        if(nitem && find_item_shown_index(nitem) != INT_MAX) {
```
GrayHatter: why?

ingvar1995: because of bug with dragging contacts. open list of online contacts, try to drag online contact to free place where offline was. offline contact and online contact will swap. so you will have offline contact in list of online contacts

PR #177, opened 2016-01-18
fixes #125

### LINKED ISSUE #125: Old "Normal" DPI is at 90%
28fc283
Not sure if it's exactly 90%, but I can't set more accurate values.

danielisaksen:
fixes issue #125, at least for me on linux, needs testing on other platforms


danielisaksen:
Would like someone to test on OS X and Android to make sure it is solved 
across all platforms.
I have tested it on two Windows machines and two Linux machines, where 
it is solved.

PR #190, opened 2016-02-01
fixed #187

### LINKED ISSUE #187: TOX ID disappears in windows
image
Tox id sometimes disappears in windows. And cannot be copied. when tox starts it save tox id in self.id_buffer and write it to debug. but later smth happens and self.id_buffer becomes empty string. memory trouble?

ingvar1995:
fixed #187

### DIFF
--- a/src/windows/main.c
+++ b/src/windows/main.c
@@ -1406,8 +1406,8 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PSTR cmd, int n
     thread(toxcore_thread, NULL);
 
     //wait for tox_thread init
-    while(!tox_thread_init) {     
-        Sleep(1);     
+    while(!tox_thread_init && !encrypted_profile) {    
+        yieldcpu(1);     
     }
 
     if (*cmd) {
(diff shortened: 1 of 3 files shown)


src/windows/main.c
```diff
@@ -1405,6 +1405,11 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PSTR cmd, int n
     //start tox thread (hwnd needs to be set first)
     thread(toxcore_thread, NULL);
 
+    //wait for tox_thread init
+    while(!tox_thread_init) {     
+        Sleep(1);
```
cebe: better use `yieldcpu` instead of `Sleep`.

GrayHatter:
Do you use a profile password?

ingvar1995:
No. I updated code for encrypted profiles

GrayHatter:
Did you test this, with the first reading I think some of this code will never run.

ingvar1995:
after update it works even with encrypted profiles.

GrayHatter:
merged into release-v0.6

PR #196, opened 2016-02-06
Av refactor [DO NOT MERGE]

GrayHatter:
This is for preview only.

Audio is done 

audio out works
audio in works
calls work
ringtone works


cebe:
@GrayHatter make fails on debian: 

```
src/xlib/main.c:1337:23: error: ‘AUDIO_KILL’ undeclared (first use in this function)
     postmessage_audio(AUDIO_KILL, 0, 0, NULL);

```

PR #204, opened 2016-02-14
More helpful messages.

jantatje:
Some of the messages in read_custom_theme() weren't particularly helpful, change them to something more descriptive.

previously was:
> Title: Make error/warning message less offensive.
> 
> uTox prints "whad de fug DDD-xx" on console if it cannot open a file, this can be deemed offensive and unprofessional. Worst case this could hinder adoption of uTox. As uTox continues running fine afterwards change it to: "warning".
> 
> While personally this wouldn't stop me from using uTox, I don't know how others feel about it. Better safe than sorry. It seems there are some more "funny" error messages in uTox, but as they weren't offensive I left them in.

### DIFF
--- a/src/theme.c
+++ b/src/theme.c
@@ -292,7 +292,7 @@ void theme_load(char loadtheme) {
         int size = sizeof("utox_theme.ini");
 
         if (len + size > 1024) {
-            puts("datapath too long, abandoning ship!");
+            debug("datapath too long, abandoning ship!\n");
             break;
         }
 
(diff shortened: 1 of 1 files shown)


GrayHatter:
if you're going to take some of the humor and whimsy out of the code, you should at least try to replace it with something helpful, in this case, the warning needs to be much more descriptive as to what that warning actually means. 

oh, btw http://www.merriam-webster.com/dictionary/fug

optimumtact:
puts("omg the file system is gone!")

jantatje:
You are right, my bad. I replaced the messages with something more helpful. As a few lines down, being unable to parse a color was deemed an error I changed the earlier one that fails the whole function to say error, too. Please have a look at it.

GrayHatter:
Can you use debug instead of puts?

jantatje:
Done.
Output looks like this now:
```
Loading custom theme: /home/jnt/.config/tox/utox_theme.ini
error: failed to open theme: No such file or directory
```

GrayHatter:
merged to release-v0.6

PR #214, opened 2016-02-25
Fix dmg script

RubenRocha:
The path was incorrect, I have corrected it to `src/cocoa/frozen_dmg.tar.bz2`


stal888:
Looks good, but please re-submit to develop.

GrayHatter:
merged into develop by hand

PR #228, opened 2016-03-04
updated Ukrainian translation

Yalvex:
Sorry i deleted the branch. It seems now all should be good.  Thanks  @cebe for finding why fails build


cebe:
:+1:

GrayHatter:
@Yalvex @cebe thanks

PR #240, opened 2016-03-11
Fix for issue 230, Error making call

master-passeli:
Patch fixing issue 230 "Error making call xlib (utox Alpha version
0.6.1)"

friendlist instanse 0 call status was not set correct on new call
pending, answer or ending cases.

This should fix audio call, might also fix video but i'm not able to test it. :(

Also function utox_audio_friend_accepted sets now call state in it so
that utox_callback_av_change_state function does not do that directly.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/grayhatter/utox/240)

### DIFF
--- a/src/utox_av.c
+++ b/src/utox_av.c
@@ -56,16 +56,14 @@ void utox_av_ctrl_thread(void *args) {
                     break;
                 }
                 case UTOXAV_INCOMING_CALL_ANSWER: {
-                    if (msg->param1) {
-                        VERIFY_AUDIO_IN();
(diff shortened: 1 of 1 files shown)


GrayHatter:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/grayhatter/utox/240#-:-KCbe3HScCV8RlHt95YX:1417088721)*

GrayHatter:
confirmed working from IRC user

PR #247, opened 2016-03-17
Make installation process more tunable (adjust pkg-config name and datarootdir)

ony:
On some platforms `pkg-config` may have cross-prefix and look like `x86_64-pc-linux-gnu-pkg-config`.

Besides that location of `datarootdir` not always is subfolder of `prefix`. In Exherbo linux filesystem layout looks like this: `/usr/x86_64-pc-linux-gnu/bin` and `/usr/share/man`. This is done intentionally to make platform-specific files placed in an individual folder for each platform while independent files can be shared in common location.

---
This change is [Reviewable](https://reviewable.io/reviews/grayhatter/utox/247)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,7 @@ endif
 
 DESTDIR ?=
 PREFIX ?= /usr/local
+DATAROOTDIR ?= $(PREFIX)/share
 
 SRC = $(wildcard src/*.c)
 HEADERS = $(wildcard src/*.h src/*/*.h langs/*.h)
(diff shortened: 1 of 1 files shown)


GrayHatter:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/grayhatter/utox/247#-:-KD4op6YqmpY4mN4SWTg:1417088721)*

Issue #233, opened 2016-03-08
Beep on new message.


viruzx:
This is just a quality of a life feature that should be very easy to implement. Just a beep on new message.

utoxxer:
Looking very forward to this one. In utox on linux with gtk environments, the panel/dock will flash to let you know you have a new message, but on qt desktops like KDE this is more faint and difficult to see. So it's hard to tell when a new message arrives.

GrayHatter:
I'm still waiting for @irungentoo to share a copy of his C music code

tr37ion:
How about bringing Ubuntu Touch message sounds to utox then? I really like utox and its dark theme. They are really nicely done.

roranicus:
If such a feature goes in, please make it optional. The reason why I use uTox is that, unlike qTox, it doesn't make my taskbar flash or otherwise gets in my face with notifications. This needs to be something that can be easily disabled.

GrayHatter:
see also #207

GrayHatter:
in develop

Issue #209, opened 2016-02-22
Password "show" button has wrong tooltip


cebe:
![bildschirmfoto von 2016-02-22 10 14 43](https://cloud.githubusercontent.com/assets/189796/13214005/32512f30-d94d-11e5-9238-9f7b1ac1babe.png)

GrayHatter:
whoopsies! Good catch, thanks.

Is that the actual alignment? It shouldn't be that far off. What DPI are you on?

cebe:
I'm at DPI "Normal (100%)" but it is similarily wrong aligned on other DPIs.

PR #250, opened 2016-03-20
minor changes

ninedotnine:
I started off trying add #233 but then I got distracted and I had a harder time understanding the audio code. I'd still like to work on that, but I've spent too much time on this today already...

For now; this branch:
- fixes the alignment problem mentioned in #209 (I was seeing this too)
- adds a non-functioning button for future "beep on message" setting
- adds a couple of French language phrases

---
This change is [Reviewable](https://reviewable.io/reviews/grayhatter/utox/250)

### DIFF
--- a/langs/fr.h
+++ b/langs/fr.h
@@ -112,6 +112,9 @@ msgstr("Périphérique de sortie audio")
 msgid(VIDEOINPUTDEVICE)
 msgstr("Périphérique d'entrée vidéo")
 
+msgid(PUSH_TO_TALK)
+msgstr("Pousser à parler")
+
 msgid(BEEP_ON_MSG)
 msgstr("Sond au message")
 
--- a/langs/fr.h
+++ b/langs/fr.h
@@ -112,6 +112,9 @@ msgstr("Périphérique de sortie audio")
 msgid(VIDEOINPUTDEVICE)
 msgstr("Périphérique d'entrée vidéo")
 
+msgid(BEEP_ON_MSG)
+msgstr("Sond au message")
+
 msgid(OTHERSETTINGS)
 msgstr("Autres paramètres")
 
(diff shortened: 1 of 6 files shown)


src/ui.c
```diff
@@ -1120,6 +1122,14 @@ void ui_set_scale(uint8_t scale) {
             .width  = UTOX_SCALE(20 )
         },
 
+        d_beep_on_msg = {
+            .type   = PANEL_DROPDOWN,
+            .x      = UTOX_SCALE(119 ),
+            .y      = UTOX_SCALE(15 ),
+            .height = UTOX_SCALE(12 ),
+            .width  = UTOX_SCALE(20 )
```
GrayHatter: please change all of these to SCALE(number * 2) utox_scale is going to be removed soon

langs/fr.h
```diff
@@ -112,6 +112,12 @@ msgstr("Périphérique de sortie audio")
 msgid(VIDEOINPUTDEVICE)
 msgstr("Périphérique d'entrée vidéo")
 
+msgid(PUSH_TO_TALK)
+msgstr("Pousser à parler")
```
Encrypt: Hum... I am not sure that it is the correct translation...
I don't know exactly what this functionality does but "to" might be in this case "pour" instead of "à".
That is to say: "Pousser pour parler" instead of "Pousser à parler".

langs/fr.h
```diff
@@ -112,6 +112,12 @@ msgstr("Périphérique de sortie audio")
...
+
+msgid(BEEP_ON_MSG)
+msgstr("Sond au message")
```
Encrypt: Here however, there is definitively a mistake. **There is no "d" at "son"!**
Moreover, this translation doesn't sound very good at all... better translations could be:
- "Sonner au message" which is slightly better but still not that good.
- "Biper au message"
- Or you may simply translate the idea, that is to say: "Notification sonore".

GrayHatter:
this seems to have gone stale.

I'm happy to merge the language if it's in its own PR, but I can't merge a half done feature.

PR #257, opened 2016-03-28
Added Russian transation for the pop-up comment of the shortcut

TotalCaesar659:
This change is [Reviewable](https://reviewable.io/reviews/grayhatter/utox/257)

### DIFF
--- a/src/utox.desktop
+++ b/src/utox.desktop
@@ -3,6 +3,7 @@ Version=1.0
 Type=Application
 Name=uTox
 Comment=A lightweight Tox client
+Comment[ru]=Нетребовательный к ресурсам клиент Tox
 TryExec=utox
 Exec=utox
 Icon=utox


src/utox.desktop
```diff
@@ -3,6 +3,7 @@ Version=1.0
 Type=Application
 Name=uTox
 Comment=A lightweight Tox client
+Comment[ru]=Нетребовательный к ресурсам клиент Tox
```
cebe: I do not think `lightweight` only refers to computer resources. I am not a native speaker but something like `легкий`, `простой`, or `несложный` would be more fitting imo.

TotalCaesar659:
OK, I just tried to satisfy it. Though, you can correct it to "Легкий клиент Tox" if you want.

GrayHatter:
@cebe thoughts on this one?

cebe:
@GrayHatter change it as `Легкий клиент Tox` and close this one?

alexeysvrv:
So, I support Russian their native language, I suggest using a combination of a `легковесный клиент` It is more accurately suited and positioned uTox

PR #279, opened 2016-04-06
attempting to make tools/android-build.sh work (WIP)

felix-salfelder:
trying to build this thing (for android.. `:|`).

first of all, i need more flexible paths, i think it make sense to pick them from the environment.
with my first commit i can run

    bash$ TOOLCHAIN_NAME=arm-linux-androideabi-4.9 SDK_PATH=~/.opt/android-sdk-linux/ \
       ANDROID_NDK_HOME=~/.opt/android-ndk-r11c/ \
       tools/android-build.sh

now, i'm getting `ls: cannot access ../openal-arm/lib/: No such file or directory`. can you tell me what this is, and where i can get it from? EDIT: found it. sorry for the noise.

thank you

---
This change is [Reviewable](https://reviewable.io/reviews/grayhatter/utox/279)

### DIFF
--- a/tools/android-build.sh
+++ b/tools/android-build.sh
@@ -1,7 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Script to cross compile utox for android from linux (and soon windows)
 #
 
+set -e
+set -x
 
 usage() {
 	echo "Cross compile script to build uTox for android from a ~unix environment."
@@ -22,6 +24,9 @@ for arg in "$@"; do
 			;;
 	esac
 done
+
+# read settings from a custom settings file. the defaults don't seem to be very useful.
+[ -f settings.android ] && source settings.android
 
 export ALIAS=utox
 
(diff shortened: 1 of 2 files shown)


felix-salfelder:
one more question: there are two similar android build scripts in `tools`. wouldn't it make sense to merge them?

GrayHatter:
https://github.com/GrayHatter/uTox/blob/android/tools/build-android.sh

This is the one i'm using now. I couldn't get the other to work.

felix-salfelder:
thanks.

both of them did not work here (mostly due to hardcoded paths that do not exist on my end), but i can give this one a try, too...

GrayHatter:
For some odd reason, android wont let you change the singing key of the build on the play store, when I learned that, I just rewrote the whole thing.

felix-salfelder:
some parts of "android-build.sh" are missing in "build-android.sh". what is `./toolchain in build-android.sh`? or, where is the toolchain supposed to be? and what is lib/arm? i don't have this here.

and: what is cpu-features.c? it's not in android-build.sh.

thank you

GrayHatter:
@felix-salfelder so I merged your changes into `android` then patched your script a bit.

now it should work for you, you just have to grab the arm libraries from build.tox.chat and `tar xf` them into `lib/arm/` and everything **should** work.

Find me here or on IRC if you need more help!

If you want to be super awesome, you could add another script that'll `curl -O` from build.tox.chat?

felix-salfelder:
> you could add another script that'll curl -O from build.tox.chat?

honestly, my personal reason for the build attempt is to turn the *source code* into an apk. so i'd rather not want to incorporate even more binaries. my idea is to line up the source repositories for the various projects and build one after the other. i use `CPPFLAGS` and `LDFLAGS` etc. (from `settings.android`) to point to the respective locations of the other packages' build artifacts.

of course (just in case my approach works), you will in principle be able to 

    curl -O stuff
    tar -xf stuff
    CPPFLAGS=-Istuff/ LDFLAGS=-Lstuff/ build_android.sh

but these lines should clearly go to a different file.

another thought: toxcore also provides build scripts, however, these merely invoke the usual build system with specialised options. it might make a lot of sense to do the same here (i.e. use the Makefile provided at top level, rather than reimplement it as a shell script).

PR #295, opened 2016-04-24
Messages queue

GrayHatter:
This change is [Reviewable](https://reviewable.io/reviews/grayhatter/utox/295)


GrayHatter:
careful, this might resend every message every time you connect to a friend...

PR #299, opened 2016-04-25
Mini friends list 

GrayHatter:
The roster can be a bit bulky in it's current shape.

I also changed the settings struct, so this might not compile on other platforms (xlib works)

---
This change is [Reviewable](https://reviewable.io/reviews/grayhatter/utox/299)

### DIFF
--- a/langs/en.h
+++ b/langs/en.h
@@ -199,6 +199,9 @@ msgstr("Away")
 msgid(STATUS_BUSY)
 msgstr("Busy")
 
+msgid(SETTINGS_UI_MINI_ROSTER)
+msgstr("Use mini contact list")
+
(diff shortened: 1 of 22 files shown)


GrayHatter:
I probably should fix the icons for users w/o an avatar, but I'm too tired for svg.c right now.

Here have some images!

### before
before

### after
after

cebe:
looks good to me, except the icon thing. This should be adjusted for users as well as groups.

cebe:
> I probably should fix the icons for users w/o an avatar, but I'm too tired for svg.c right now.

here you are https://github.com/GrayHatter/uTox/pull/304 :)

PR #374, opened 2016-06-25
Update cygwin-compile.sh

endoffile78:
This change is [Reviewable](https://reviewable.io/reviews/grayhatter/utox/374)

### DIFF
--- a/tools/cygwin-compile.sh
+++ b/tools/cygwin-compile.sh
@@ -88,9 +88,8 @@ LIBVPX="./lib/vpx"
 LIBOPUS="./lib/opus"
 LIBOPENAL="./lib/openal"
 
-GIT_V=`git describe --abbrev=8 --dirty --always --tags`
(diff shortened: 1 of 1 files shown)


GrayHatter:
I've updated `make` to work on cygwin as well. I'll merge this if you first try `CC=x86_64-w64-mingw32-gcc make` It should greatly speedup your compile times if you're only changing small parts of code.

endoffile78:
@GrayHatter I tried that before and I couldn't get it to work so I decided to use the cygwin-compile script. I have been having trouble getting it to link openal and libsodium. They are installed in cygwin and I have the mimgw versions from cygwin too.

GrayHatter:
merged, thanks mate!

I followed the instructions on toxcore for cross compiling everything, and with `make install` on all of them from within cygwin, everything seems to work for me. But I may have done something strange that I've forgotten about.

Hope this works for you

PR #397, opened 2016-07-23
Converted all yes / no dropdowns to switches

Encrypt:
Here are all the switches corresponding to the previous yes / no dropdowns!

switches

---
This change is [Reviewable](https://reviewable.io/reviews/grayhatter/utox/397)


xaizek:
I'm missing yes/no dropdowns... Not just because switches are counter intuitive by design regardless of their color or label, but also because this looks weird if you make the window wide (hard to tell which labels correspond to which switches). Also checkout "Audio/Video" pane, which now has "Group Notifications" *between* labels and corresponding switches.

GrayHatter:
I have no plans to change it back. If you hover your mouse over the switch, each should pop up a tooltip for the setting it's going to change. Finally the audio notification for groups is only in the development version because I haven't made time to move it yet.

PR #414, opened 2016-08-11
pass LIBS to the linker

felix-salfelder:
please allow to inject libs, in case pkg-config fails.

pkgconfig seems broken all the time here (don't ask me why). in this situation, this patch re-enables the build. `make CFLAGS=this LIBS=that`.

(this is related to https://github.com/qTox/qTox/issues/3374, other tox clients also use pkg-config)

---
This change is [Reviewable](https://reviewable.io/reviews/grayhatter/utox/414)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ all: utox
 
 utox: $(OBJ) $(OS_OBJ) $(TRAY_OBJ)
 	@echo "  LD    $@"
-	@$(CC) $(CFLAGS) -o $(OUT_FILE) $(OBJ) $(OS_OBJ) $(TRAY_OBJ) $(LDFLAGS)
+	@$(CC) $(CFLAGS) -o $(OUT_FILE) $(OBJ) $(OS_OBJ) $(TRAY_OBJ) $(LDFLAGS) $(LIBS)
 
 install: utox
 	mkdir -p $(DESTDIR)$(PREFIX)/bin
--- a/Makefile
+++ b/Makefile
@@ -17,8 +17,22 @@ LDFLAGS += -pthread -lm
 # also link against these
 LIBS =
 
+ifeq ($(UNAME_S), Linux)
+	PKG_CONFIG = pkg-config
+else ifeq ($(UNAME_O), Cygwin)
+	PKG_CONFIG = x86_64-w64-mingw32-pkg-config
+else
+	PKG_CONFIG = false
+$(error unsupported platform)
+endif
+
 ifeq ($(FILTER_AUDIO), 1)
+ifeq ($(shell $(PKG_CONFIG) --exists filteraudio && echo yes), yes )
 	DEPS += filteraudio
+else
+$(warning pkg-config does not know about filteraudio, this might not work)
+$(error consider setting FILTER_AUDIO=0)
+endif
 	CFLAGS += -DAUDIO_FILTERING
 endif
 
@@ -45,8 +59,6 @@ ifeq ($(UNAME_S), Linux)
 		CFLAGS += -DNO_DBUS
 	endif
 
-	PKG_CONFIG = pkg-config
-
 	CFLAGS += $(shell $(PKG_CONFIG) --cflags $(DEPS))
 
 	LDFLAGS += -lresolv -ldl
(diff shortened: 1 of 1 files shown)


GrayHatter:
Any reason you can't set LDFLAGS? 

Or, What's wrong with pkg-config? Maybe we should fix that.

felix-salfelder:
LDFLAGS is for another purpose, and already set. if I override this, it breaks on the other end.
putting LIBS into LDFLAGS somehow feels like reusing CFLAGS for the OBJS...

i don't have a better explanation, c.f. [gnu make docs](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html#Implicit-Variables) for a reference. they recommend "Libraries (-lfoo) should be added to the LDLIBS variable instead." LDLIBS or LIBS, whichever you like best.

> Or, What's wrong with pkg-config? Maybe we should fix that.

i am not sure yet. it seems that pkgconfig fails and the makefile silently ignores the error. this might not be visible in most situations... maybe i can fix it (some day).

thanks

GrayHatter:
Ok, I'll merge this if you add the var to the top, and a comment as to why is exists

GrayHatter:
Sorry, I meant just a blank assignment, and comment that it's needed on some systems where pkg-config is broken. Don't touch LDFLAGS at all

felix-salfelder:
ok, ive got something... this is a bit of a workaround

the original problem is, that DEPS is compiled first and the passed to pkg-config. it would be much easier (slightly more repetitive) to fetch them one by one.

proper checks for all of them (i only added a filteraudio check) will end up repetitive as well...

GrayHatter:
I'm going to push a branch with a new Makefile I've been working on. Take a look at that too, and see if there's anything you might want to add in.

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[Makefile, line 30 \[r3\]](https://reviewable.io:443/reviews/grayhatter/utox/414#-KOvSfXJbaujlQu3G_us:-KOvSfXK68kymRdSfPGr:bei8qx6) ([raw file](https://github.com/grayhatter/utox/blob/af508b4a63891ed945e1a8ae06a9ec25c97ea055/Makefile#L30)):*
> ```Makefile
> 
> ifeq ($(FILTER_AUDIO), 1)
> ifeq ($(shell $(PKG_CONFIG) --exists filteraudio && echo yes), yes )
> ```

Need indentation.

---

*[Makefile, line 34 \[r3\]](https://reviewable.io:443/reviews/grayhatter/utox/414#-KOvT0OqKtg4-t_SsjGS:-KOvT0OqKtg4-t_SsjGT:bjh1y42) ([raw file](https://github.com/grayhatter/utox/blob/af508b4a63891ed945e1a8ae06a9ec25c97ea055/Makefile#L34)):*
> ```Makefile
> else
> $(warning pkg-config does not know about filteraudio, this might not work)
> $(error consider setting FILTER_AUDIO=0)
> ```

Might want to move this test up a few sections, and actually set filter audio

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/grayhatter/utox/414#-:-KOvT8huqV97KaSDseR7:bfdgqnr)*

GrayHatter:
@felix-salfelder still working on this one?

felix-salfelder:
yes, and no. thanks for asking, i forgot to report back. my workaround is included [here](https://anonscm.debian.org/git/pkg-voip/utox.git/tree/debian/rules).

one more problem i encountered is, that too many libraries are listed in $(LDFLAGS). debhelper complained about utox being linked against libsodium and libopus (and debhelper was right). my workaround takes care of that, but i havent got a good solution to present here.

thinking about it. using LIBS, then splitting LIBS into LIBS_THIS and LIBS_THAT would at least simplify the workaround.

e.g. i could then do

    $ make [..] LIBS_OPUS= LIBS_SODIUM=

to avoid linking to these.

what are your plans? autoconf would certainliy be great `:)`.

GrayHatter:
I'm looking into either cmake (everyone says it's great, and has switched)
or autoconf because I hate cmake (only because I don't understand it, and
haven't yet read a tutorial that doesn't suck)

felix-salfelder:
of the autotools alternatives, cmake is likely one of the best. imo, it's still not worth the trouble. mostly a shift from "understand how things are implemented" to "implement it yourself, if you want".

eventually cmake is easy for the developer, and often a nightmare for the user. autotools can be the other way around, but it does not have to be. let me know, if you need help or testing.

GrayHatter:
closed until I can learn cmake

PR #415, opened 2016-08-11
Add wait() to xlib to reap terminated child procs

somercet:
uTox with Xlib does not reap terminated child processes, such as the xdg-open command called by uTox's openurl(). The following is not a "real" fix: child pids aren't tracked, no messages print on child errors but it does:
1. free zombies without glib or global child pid tables, and
2. reduce outstanding zombies to one at any time,
3. without blocking and
4. without looping/polling.

---
This change is [Reviewable](https://reviewable.io/reviews/grayhatter/utox/415)

### DIFF
--- a/src/xlib/main.c
+++ b/src/xlib/main.c
@@ -218,6 +218,8 @@ void openurl(char_t *str)
     if(!fork()) {
         execlp(cmd, cmd, str, (char *)0);
         exit(127);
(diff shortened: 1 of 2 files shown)


GrayHatter:
can you rebase/merge develop?

I'll take a closer look then, and hopefully merge soon after.

PR #425, opened 2016-09-03
Fix disabled hover color in themes

tsudoko:
Elements which use this color will be no longer black.

---
This change is [Reviewable](https://reviewable.io/reviews/grayhatter/utox/425)

Issue #421, opened 2016-08-30
Switch UI element visually broken


davidak:
![bildschirmfoto 2016-08-31 um 01 43 08](https://cloud.githubusercontent.com/assets/91113/18111003/5594a624-6f1c-11e6-9ed2-44c260631745.png)

---

v0.9.8
OS X 10.10.5

GrayHatter:
What's broken about it?

davidak:
Well, it looks broken. White element on white background.

Maybe a design is missing or some graphic is not loaded.

GrayHatter:
what should it look like?

davidak:
A Designer should decide what fit's the UI of uTox.

There are a lot of designs of switches:

![](https://www.b4x.com/android/forum/proxy.php?image=http%3A%2F%2Fwww.visualnet.inf.br%2Fimagens%2Fb4a%2FUISwitchTOP.jpg&hash=25864df58e02287ccd0db3737e1beada)

If i should suggest something:

We have already this rounded buttons:

<img width="233" alt="bildschirmfoto 2016-09-01 um 12 20 54" src="https://cloud.githubusercontent.com/assets/91113/18164173/c23d9ace-703e-11e6-98b5-8d16bab51323.png">

(that are too big for the text btw)

so something like this would fit:

![serveimage](https://cloud.githubusercontent.com/assets/91113/18164199/efb52e54-703e-11e6-80de-61f98dce5d5c.jpeg)

When the setting is to activate a feature, a ✓ with green background is nice, not activated should be grey instead of red since red means danger.

Encrypt:
The design of the switches that you are suggesting looks good @davidak!

I'll try to change them once I have time (i.e. once my MSc is over... that is to say in a few weeks).

davidak:
@Encrypt Very good.

Can someone change the color of the white thing to dark grey until that?

GrayHatter:
I like your suggestion ALOT A LOT! @Encrypt I hope you beat me to it, but if not I'll see if I can find time this weekend

GrayHatter:
half fixed

GrayHatter:
fixed a while ago by @tsudoko, and it looks awesome! thanks mate!

PR #426, opened 2016-09-03
Fix switches

tsudoko:
Sort of depends on #425
See #421 for more information, still WIP

image

Current progress:
- [x] ~~draw two vertical lines on top of the inner part~~
- [x] use disabled color when off
- [x] make the inner part smaller
- [x] make the outer part's edges rounded
- [x] add `on`/`off` icons (not sure if needed)
- [x] de-hardcode toggle positions
- [x] de-hardcode icon styles
- [x] ~~make the `on` icon not inverted~~ see #429

---
This change is [Reviewable](https://reviewable.io/reviews/grayhatter/utox/426)

### DIFF
--- a/src/main.h
+++ b/src/main.h
@@ -254,6 +254,7 @@ typedef enum {
     BM_SBUTTON,
 
     BM_SWITCH,
+    BM_SWITCH_ICON,
 
     BM_CONTACT,
     BM_CONTACT_MINI,
--- a/src/main.h
+++ b/src/main.h
@@ -254,7 +254,7 @@ typedef enum {
     BM_SBUTTON,
 
     BM_SWITCH,
-    BM_SWITCH_ICON,
+    BM_SWITCH_INNER,
 
     BM_CONTACT,
     BM_CONTACT_MINI,
--- a/src/main.h
+++ b/src/main.h
@@ -254,7 +254,7 @@ typedef enum {
     BM_SBUTTON,
 
     BM_SWITCH,
-    BM_SWITCH_INNER,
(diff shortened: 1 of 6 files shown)


GrayHatter:
So far, I haven't seen anything worrying, but you're starting to get close to options overload.

I like everything so far, but I don't want to add too many decisions to switches, because they're supposed to be an easy UI element.

---

Reviewed 6 of 6 files at r1, 6 of 6 files at r2.
Review status: 5 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/426#-:-KR8cTbKf-SPnHz7xiEh:bsty52l)*

GrayHatter:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/426)*

tsudoko:
Do you think some graphics should be hardcoded? I can leave on/off icons as the only changeable graphics if the number of options is an issue.

GrayHatter:
I thought I should say yes, even started writing a comment as such. But
before I could hit submit, it always seems like a better idea to have the
options.  Some toggles (in the future should/could be styled to be ON when
on the left, or red when you're specifically blocking some actions) In
those cases they way you're writing it makes much more sense.

tsudoko:
What about the outer part and the toggle? I can't think of any case where using different outer/toggle graphics only for some switches would make sense.

GrayHatter:
@tsudoko so I'm thinking of merging this ≈soon unless you want to add a few more commits/LOC?

PR #442, opened 2016-10-07
Add header guards

endoffile78:


src/android/main.h
```diff
@@ -1,3 +1,6 @@
+#ifndef ANDROID_MAIN_H
+#define ANDROID_MAIN_H
```
GrayHatter: for any main.h we should 
```c
#if defined(MAIN_H) && !defined(MAIN_H_ANDROID)
#error we should never include main from different platforms
...
#define MAIN_H
#define MAIN_H_ANDROID
```

Suggestion, not mandatory. I'm open to other thoughts.

endoffile78: @GrayHatter I think thats a good idea I'll add that.

GrayHatter:
I'll merge this sometime tomorrow, I just want to give anyone else a chance to comment, like @stal888 or @tsudoko or anyone really.

The next step will be to start dropping includes from src/main.h and instead actually adding them ONLY to the file that need them.

PR #462, opened 2016-11-03
Updated links in CHANGELOG.md to point to new repo

robinlinden:
Repo urls updated from https://github.com/GrayHatter/uTox/ to https://github.com/uTox/uTox/


robinlinden:
Just noticed that there's an open PR that fixes this and more, so never mind.

Issue #439, opened 2016-10-01
The "Show" button is on the title it meant to be next to


SeventhSonOfASeventhSon:
I have found that on Ubuntu GNOME 16.04.1 with GNOME 3.20 that on uTox `0.9.7` that the `Show` button in the Profile tab in the Preferences is ontop of the title it is meant to be next to:

![show_button_on_password_label](https://cloud.githubusercontent.com/assets/11575271/19014893/fe833d6c-87f0-11e6-8fbe-1b0e6386fcce.png)

PR #482, opened 2016-11-16
Clicking "show profile password" button hides it.

### LINKED ISSUE #440: "Show" button remains even when it has performed its function
I have noticed that when click the "Show" button here that it remains even though it has already revealed what it was there to reveal:

show_button_does_not_change

I believe that it would be better for this button to either go when it has fulfilled its function, or become a "Hide" button which actually does something because pressing "Show" again does nothing.

I am running Ubuntu GNOME 16.04.1 with GNOME 3.20 and uTox version `0.9.7`.

robinlinden:
Fixes #440 and #439

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/482)

### DIFF
--- a/src/ui.c
+++ b/src/ui.c
@@ -256,7 +256,7 @@ void ui_set_scale(uint8_t scale) {
 
     /* Profile              */
     CREATE_BUTTON(copyid, SCALE(66), SCALE(106), BM_SBUTTON_WIDTH, BM_SBUTTON_HEIGHT);
-    CREATE_BUTTON(show_password_settings, SCALE(130), SCALE(206), BM_SBUTTON_WIDTH, BM_SBUTTON_HEIGHT);
+    CREATE_BUTTON(show_password_settings, SCALE(145), SCALE(206), BM_SBUTTON_WIDTH, BM_SBUTTON_HEIGHT);
     CREATE_BUTTON(lock_uTox, SCALE(10), SCALE(260), BM_SBUTTON_WIDTH, BM_SBUTTON_HEIGHT);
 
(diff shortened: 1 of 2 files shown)


GrayHatter:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[src/ui/buttons.c, line 390 at r1](https://reviewable.io:443/reviews/utox/utox/482#-KWfsTNSjoMldUA94NtZ:-KWfsTNSjoMldUA94Nt_:b-8flzyk) ([raw file](https://github.com/utox/utox/blob/3cc53b15c567f9e252a414d013c3894fa7974e85/src/ui/buttons.c#L390)):*
> ```C
> 
> static void button_show_password_settings_onpress(void) {
>     button_show_password_settings.disabled = 1;
> ```

should be true/false here and there

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/482#-:-KWfsZRiglmsFX3s0HAD:bnfp4nl)*

robinlinden:

---

*[src/ui/buttons.c, line 390 at r1](https://reviewable.io:443/reviews/utox/utox/482#-KWfsTNSjoMldUA94NtZ:-KWftsXhXk23gu-eobRH:b-wpqlms) ([raw file](https://github.com/utox/utox/blob/3cc53b15c567f9e252a414d013c3894fa7974e85/src/ui/buttons.c#L390)):*
<details><summary><i>Previously, GrayHatter (Gregory Mullen) wrote…</i></summary>
> should be true/false here and there
</details>
Will do this together with all other UI stuff that needs to be switched over to true/false in a separate PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/482)*

PR #535, opened 2016-12-12
Fix various warnings

GrayHatter:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/535)


robinlinden:
The commits in this PR are a subset of the commits in #536. Want to drop the refactor commit from this PR and the two warning commits from #536?

PR #666, opened 2017-01-24
Decrease minimum width.

### LINKED ISSUE #663: Min window size limit
uTox window's minimum size is limited depending on DPI and it's OK, but imo it limited too much, because now at my normal DPI 110% it consumes almost whole my screen, which is nearly unusable in non-fullscreen mode.

tox

I don't know this is bug or feature (or it's fix for #581), but it really frustrates me.

robinlinden:
I like to have the uTox window as small as possible along the horizontal axis. A problem with this is that it cuts off the end of the nospam message. In my opinion that's fine as you can expand the window to read it the one time you need to.

I would like to make it even smaller horizontally, but at that point the Tox ID starts getting cut off and users might be confused by that.

nospam

Fixes #663

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/666)

### DIFF
--- a/src/main.h
+++ b/src/main.h
@@ -50,8 +50,8 @@ typedef uint8_t *UTOX_IMAGE;
 
 #define BORDER 1
 #define CAPTION 26
-#define MAIN_WIDTH 1000
-#define MAIN_HEIGHT 600
+#define MAIN_WIDTH 800
+#define MAIN_HEIGHT 500
 
 //  fixes compile with apple headers
 /*** This breaks both android and Windows video... but it's needed to fix complation in clang (Cocoa & asan)


GrayHatter:
we can also make that two lines
eg, like the password warning

GrayHatter:
also, I fixed the GUI hang bug so we can make it much smaller as well

GrayHatter:
:lgtm:

or we can make it smaller now too

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/666#-:-KbKD-8Tmrkq-a1HnERt:b-c7jmj)*

GrayHatter:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/666)*

robinlinden:
You'd hide most of the settings tabs before the client gets to thin that the UI chat hang problem comes into effect. This is the smallest width possibly without starting to hide the Tox ID. 

We might want to make this change and then make a larger PR adding ellipses when part of the Tox ID is covered, maybe reduce the padding of the settings tabs, wrap the warning texts, and other things that would enable us to make the interface less wide without breaking things.

GrayHatter:
> UI chat hang problem comes into effect. This is the smallest width possibly without starting to hide the Tox ID.

This is fixed already. (About a week ago.) 

So the only question left is; are we okay with obsecuring the ToxID at all? I'm okay with that if we can find a reasonable min? Perhaps the min width for the settings tabs?

Then: OOS maybe we should detect that width, and user shorter names?

:lgtm:

robinlinden:
Yeah, what I meant was that we probably couldn't encounter the UI hang without breaking all settings menus first. 

Min width that the settings tabs is  700 < 750px iirc. (That's what I was going to do originally before I noticed that that cut off the end of the Tox ID.) 

Adding the logic for all of this is OOS for this PR. Let's merge this for 0.12.2 and open a new one with a lower width and the UI fixes.

fcore117:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/666)*

endoffile78:
:lgtm:

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/666#-:-KbWtEhsHofTLQHZDkoA:bnfp4nl)*

PR #731, opened 2017-02-12
Update CMakeLists.txt

Axaprj:
openal-soft generates always OpenAL32 when builds for Windows.

https://cmake.org/cmake/help/v3.5/variable/WIN32.html
>WIN32
>True on Windows systems, including Win64.
>Set to true when the target system is Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/731)

### DIFF
--- a/src/av/CMakeLists.txt
+++ b/src/av/CMakeLists.txt
@@ -7,7 +7,11 @@ add_library(utoxAV STATIC
     )
 
 if(NOT APPLE)
+if(WIN32)
+    target_link_libraries(utoxAV OpenAL32)
+else()
     target_link_libraries(utoxAV openal)
+endif()
 endif()
 
 if(FILTER_AUDIO)
--- a/src/av/CMakeLists.txt
+++ b/src/av/CMakeLists.txt
@@ -6,12 +6,11 @@ add_library(utoxAV STATIC
     video.c
     )
 
-if(NOT APPLE)
-if(WIN32)
+if(APPLE)
+elseif(WIN32)
     target_link_libraries(utoxAV OpenAL32)
 else()
     target_link_libraries(utoxAV openal)
-endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


src/av/CMakeLists.txt
```diff
@@ -7,8 +7,12 @@ add_library(utoxAV STATIC
     )
 
 if(NOT APPLE)
+if(WIN32)
+    target_link_libraries(utoxAV OpenAL32)
+else()
```
cebe: this could be an `elseif(NOT APPLE)` to make it more readable.

cebe:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/731#-:-KcnsVM4hF9NbziyRMau:bnfp4nl)*

robinlinden:
We need to make changes to Travis, the utox.io builds, and the build.tox.chat builds when we merge this. Currently all of the above rename OpenAL32 to openal to fix this bug. This is a nicer solution and should be merged.

GrayHatter:
just change the `mv` command in `windows-before.sh` to `cp`

GrayHatter:
per IRC, someone on the devel team needs to write the fix for this. 

My suggestion is that we pre-patch, and then merge this on top of that. (Or maybe we can then get @Axaprj to rebase on that patch)

GrayHatter:
@Axaprj if you can rebase this on develop, it should pass travis, and we can merge this.

Axaprj:
we can easy avoid this fix by adding into building script following:
> ln -s /usr/lib/libOpenAL32.a /usr/lib/libopenal.a

I don't know which solution will be better...

endoffile78:
@Axaprj we have fixed the build script. Can you rebase please.

PR #746, opened 2017-02-16
Don't leak memory if realloc fails.

robinlinden:
Doesn't really matter at this point, but maybe in the future, we don't want to exit() uTox when groupchats can't realloc.

Also adds more error-checks.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/746)

### DIFF
--- a/src/groups.c
+++ b/src/groups.c
@@ -174,30 +174,40 @@ void group_peer_name_change(GROUPCHAT *g, uint32_t peer_id, const uint8_t *name,
         memcpy(old, peer->name, peer->name_length);
         size = snprintf((char *)msg, TOX_MAX_NAME_LENGTH, "<- has changed their name from %.*s",
                         (int)peer->name_length, old);
-        peer = realloc(peer, sizeof(GROUP_PEER) + sizeof(char) * length);
 
-        if (peer) {
-            peer->name_length = utf8_validate(name, length);
-            memcpy(peer->name, name, length);
-            g->peer[peer_id] = peer;
-            pthread_mutex_unlock(&messages_lock);
-            group_add_message(g, peer_id, msg, size, MSG_TYPE_NOTICE);
-            return;
+        GROUP_PEER *new_peer = realloc(peer, sizeof(GROUP_PEER) + sizeof(char) * length);
+
+        if (new_peer) {
+            peer = new_peer;
         } else {
-            LOG_ERR("Groupchat", " couldn't realloc for group peer name!");
+            free(peer);
+            LOG_FATAL_ERR(40, "Groupchat", "Couldn't realloc for group peer name!");
             exit(40);
         }
+
+        peer->name_length = utf8_validate(name, length);
+        memcpy(peer->name, name, length);
+        g->peer[peer_id] = peer;
+        pthread_mutex_unlock(&messages_lock);
+        group_add_message(g, peer_id, msg, size, MSG_TYPE_NOTICE);
+        return;
     } else if (peer) {
         /* Hopefully, they just joined, because that's the UX message we're going with! */
-        peer = realloc(peer, sizeof(GROUP_PEER) + sizeof(char) * length);
-        if (peer) {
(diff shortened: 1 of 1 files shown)


cebe:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/groups.c, line 185 at r1](https://reviewable.io:443/reviews/utox/utox/746#-Kd3MOoBtxIeEUc3TsyQ:-Kd3MOoCpFSn10_4JKY9:b-adauw5) ([raw file](https://github.com/utox/utox/blob/8022bf1783c1f085749643343761de85f6ae497a/src/groups.c#L185)):*
> ```C
>             free(peer);
>             LOG_ERR("Groupchat", "Couldn't realloc for group peer name!");
>             exit(40);
> ```

shouldn't this be LOG_FATAL_ERR with exit code then? Also how do you decide for the 40, random?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/746)*

robinlinden:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/groups.c, line 185 at r1](https://reviewable.io:443/reviews/utox/utox/746#-Kd3MOoBtxIeEUc3TsyQ:-Kd3PX3ObThy4luWdm4Q:b-yoooo9) ([raw file](https://github.com/utox/utox/blob/8022bf1783c1f085749643343761de85f6ae497a/src/groups.c#L185)):*
<details><summary><i>Previously, cebe (Carsten Brandt) wrote…</i></summary><blockquote>

shouldn't this be LOG_FATAL_ERR with exit code then? Also how do you decide for the 40, random?
</blockquote></details>

You're right. We should also fix LOG_FATAL_ERR so it does its thing even if the verbosity level is 0. Right now it's just a worse LOG_ERR since we have to write the exit code twice.

Right now we don't have a system for error codes. A TODO is to add them to the debug header.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/746)*

endoffile78:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/746)*

GrayHatter:
:lgtm: 

---

Reviewed 1 of 1 files at r2.

---

*[src/groups.c, line 185 at r1](https://reviewable.io:443/reviews/utox/utox/746#-Kd3MOoBtxIeEUc3TsyQ:-Kd4a1ns5K8xN50jqYe5:buqx099) ([raw file](https://github.com/utox/utox/blob/8022bf1783c1f085749643343761de85f6ae497a/src/groups.c#L185)):*
<details><summary><i>Previously, robinlinden (Robin Lindén) wrote…</i></summary><blockquote>

You're right. We should also fix LOG_FATAL_ERR so it does its thing even if the verbosity level is 0. Right now it's just a worse LOG_ERR since we have to write the exit code twice.

Right now we don't have a system for error codes. A TODO is to add them to the debug header.
</blockquote></details>

We should use `FATAL_ERR()` then.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/746#-:-Kd4a9KeQWHfOSXpgB9A:bg7igkv)*

robinlinden:

---

*[src/groups.c, line 185 at r1](https://reviewable.io:443/reviews/utox/utox/746#-Kd3MOoBtxIeEUc3TsyQ:-Kd6pSyRlb8P0I8EZcAl:b-896fix) ([raw file](https://github.com/utox/utox/blob/8022bf1783c1f085749643343761de85f6ae497a/src/groups.c#L185)):*
<details><summary><i>Previously, GrayHatter (Gregory Mullen) wrote…</i></summary><blockquote>

We should use `FATAL_ERR()` then.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/746)*

PR #723, opened 2017-02-07
Allow changing nospam manually

endoffile78:
Fixes #708 

This change is [Reviewable](https://reviewable.io/reviews/utox/utox/723)

PR #807, opened 2017-04-09
Fix double free in edit_do on friend_free()

GrayHatter:
closes #723

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/807)

### DIFF
--- a/src/ui/edit.c
+++ b/src/ui/edit.c
@@ -5,6 +5,7 @@
 #include "scrollable.h"
 #include "text.h"
 
+#include "../debug.h"
 #include "../macros.h"
 #include "../main.h"
 #include "../settings.h"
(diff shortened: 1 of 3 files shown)


src/ui/edit.c
```diff
@@ -360,16 +366,10 @@ void edit_do(EDIT *edit, uint16_t start, uint16_t length, bool remove) {
     if (edit->history_cur != edit->history_length) {
         uint16_t i = edit->history_cur;
         while (i != edit->history_length) {
-            free(edit->history[i]);
-            i++;
+            free(edit->history[i++]);
```
robinlinden: Any particular reason you're not setting this to NULL after freeing it here too?

robinlinden:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/807#-:-KhJqPA7iJwvKZaNNRLY:bnfp4nl)*

src/ui/edit.c
GrayHatter: Yes, because this function shouldn't be able to double free without a threading issue I'd like to know about.

Where as the friend_free could be called from any thread, so there's reason to leave it in a safe state.

endoffile78:
:lgtm:

---

Reviewed 2 of 3 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/807#-:-KhKp_MM8liETCYuH7GX:bnfp4nl)*

PR #902, opened 2017-05-04
Add linux build that uses gcc

jhert0:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/902)


GrayHatter:
@endoffile78 let's disable tests for now, and see what happens

GrayHatter:
@endoffile78 all builds fail

Issue #979, opened 2017-07-28
"redefinition of typedef"; ie C99 vs C11


JamesC1:
I'm attempting to build uTox on a Mac running OSX 10.6.8 and gcc 4.2.1.

It appears to support C99 by default, which disallows repeating typedefs:
https://stackoverflow.com/questions/8594954/repeated-typedefs-invalid-in-c-but-valid-in-c/

eg:
/Users/jamescone/buildTrees/uTox/versions/pull-default/uTox/src/av/../flist.h:7: error: redefinition of typedef ‘GROUPCHAT’
/Users/jamescone/buildTrees/uTox/versions/pull-default/uTox/src/av/utox_av.h:53: error: previous declaration of ‘GROUPCHAT’ was here

If you-all want to support C99, then I can give you a pull request for a change that puts typedef'd incomplete structures into one separate header file per source directory, protected with ifdefs in the usual way.

So, do you want to support C99?

robinlinden:
If that's the only thing that makes uTox incompatible with C99 then a PR
would be nice.

We've made the decision to switch to C11, but for as long as we're not
using any features from there and keeping uTox compatible wouldn't make the
code more complicated I see no reason to break compatability.

JamesC1:
Hello Robin,

This is enough change to make av/utox_av.c compile.
https://github.com/JamesC1/uTox/commit/73119cdc23210cff4ea3437164277f45740178da

The 75% of the file after the #if 0 gives you an idea of the number of remaining duplicated typedefs.

The Cocoa notes:
https://github.com/uTox/uTox/blob/v0.15.0/docs/COCOA.md
suggest that more substantive work will also be required.

If you're good with this level of change, then I will continue.

Regards,
James.

JamesC1:
Hello Robin,

My PR is here:
https://github.com/uTox/uTox/pull/980

Simply to compile the parts of uTox which are not specific for OSX 10.7, it seems that the typedefs are the only problem.

robinlinden:
I'm traveling without access to a PC until the end of week, but I had a
quick look anyway. Prefer to make multiple headers to one monolithic one.
E.g. chatlog_fwd.h for the forward declares related to chatlogs. The plan
is to separate the UI from all logic so a monolith wouldn't work then. I'll
do a more thorough review when I'm back this weekend.

JamesC1:
I thought about one file of forwards per directory.  Knowing what I know now:
- I don't know how to name the file for src/
[...]

PR #980, opened 2017-07-30
Typedefs for c99 compilers

JamesC1:
Further to:
https://github.com/uTox/uTox/issues/979
please find attached a set of commits which (with some hacks to the cmake, which I'm still playing with, and don't link yet) appear to allow me to build all of the source for an X11 version of uTox.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/980)

### DIFF
--- a/src/android/gl.h
+++ b/src/android/gl.h
@@ -9,7 +9,7 @@
 #define _GNU_SOURCE
 #include <EGL/egl.h>
 
-typedef struct native_image NATIVE_IMAGE;
+#include "../typedefs.h"
 
 typedef struct {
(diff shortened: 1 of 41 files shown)


robinlinden:
Per IRC, this needs to be split into multiple files before it can be merged. I'll get to it eventually if no other contributor does.

GrayHatter:
I don't want a `typedef.h` file. But because this needs to be broken into smaller commits for review, you said you're still working with cmake, and there's been no update in 21 days. I'm going to close this.

Anyone can reopen it for discussion or to continue committing.

PR #1066, opened 2017-09-23
tray icon fix in d505cb8f was not correct, cleanup

termac:
Commit d505cb8f tried to read the icon_size directly from
_binary_icons_utox_128x128_png_size which is wrong.
The way it was done before via &_binary_icons_utox_128x128_png_size is
the correct way to do it.

Unfortunately this is not working anymore on systems with PIE (position
independent executable) enabled by default. This causes the loader to
relocate the absolute symbol _binary_icons_utox_128x128_png_size, which
defeats the purpose of the symbol being absolute.
(See discussion here: [1])

A proper fix would be not to use ld to generate the object file but to
serialize the png file to C code via a helper programm during the build
process. I will look into that.

For now just calculate the icon size from the symbols
_binary_icons_utox_128x128_png_start and
_binary_icons_utox_128x128_png_end.

[1] https://www.reddit.com/r/C_Programming/comments/71qx7t/reading_the_size_of_linked_binary_object_file/)

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1066)

### DIFF
--- a/src/xlib/tray.c
+++ b/src/xlib/tray.c
@@ -14,7 +14,7 @@
 
 // Converted to a binary and linked at build time
 extern uint8_t _binary_icons_utox_128x128_png_start;
-extern size_t  _binary_icons_utox_128x128_png_size;
+extern uint8_t _binary_icons_utox_128x128_png_end;
 
 static void send_message(Display *dpy, /* display */
                   Window w, /* sender (tray window) */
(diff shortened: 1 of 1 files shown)


robinlinden:
Thanks for fixing it! :lgtm_strong:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/1066#-:-KuibJR0gSilKIV5cbs0:b-ih26ua)*

redmanmale:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/1066#-:-KuigDmHJ03HZdgaeCgn:bnfp4nl)*

PR #1103, opened 2017-10-25
Add utox_preinit()

endoffile78:
This is useful for `settings.debug_file` right now its set in `parse_args`. It really should be set before because there could be errors before `parse_args` is called which would result in no output.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1103)

### DIFF
--- a/src/android/main.c
+++ b/src/android/main.c
@@ -595,6 +595,8 @@ static void utox_andoid_input(AInputQueue *in_queue, AInputEvent *event) {
 }
 
 static void android_main(struct android_app *state) {
+	utox_preinit();
+
     utox_init();
 
     theme_load(THEME_DEFAULT);
--- a/src/cocoa/main.m
+++ b/src/cocoa/main.m
@@ -516,6 +516,8 @@ int main(int argc, char const *argv[]) {
     int8_t set_show_window;
     bool   skip_updater;
 
+    utox_preinit();
+
     utox_init();
 
     settings.window_width  = DEFAULT_WIDTH;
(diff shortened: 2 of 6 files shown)


GrayHatter:
No, lets refactor init to make sure it can be called before parse_args. 

@robinlinden @cebe anyone remember why parse_args needed to be called first? If so that's the problem we should fix

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/1103#-:-KxJVvHPzcFsFcRM5r6A:b-ndusbh)*

cebe:
not sure I understand what you mean, settings.debug_file must be set before parse_args because parse_args will overwrite it, if you specify a file.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/utox/utox/1103#-:-KxKVn3ALEb-s5q7M4oX:b-wjptog)*

endoffile78:
@GrayHatter I think `utox_init` is called before `parse_args` because of the splashcreen check. I agree though it would be beter to just change `utox_init`

@cebe There could be an error message before `parse_args` is called meaning there would be no output. For example right now if this check fails:

```c
    if (!XInitThreads()) {
        LOG_FATAL_ERR(EXIT_FAILURE, "XLIB MAIN", "XInitThreads failed.");
    }
```

then there would be no output it would look like uTox just closed.

cebe:
I understand the change, I did not understand what @GrayHatter said.

GrayHatter:
closing, preinit is the wrong way to go about it.... refactor pending... eventually...

PR #1181, opened 2018-02-14
Allow passing libraries to make_test that need to be linked to the

jhert0:
test

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1181)


robinlinden:
The proper thing would be to, instead of doing things like `#include "../src/updater.c` to test the updater, you do something like `target_link_library(test_updater library_containing_updater)`, and you'd get all dependencies and header files automagically.

Basically, we'd have to rewrite our main CMakeLists to create a utox_core library or something similar.

GrayHatter:
@robinlinden meson when?

robinlinden (approved):
This is better than it was before.

avoidr:
@endoffile78 please rebase and please give some sort of summary in commit message of why your shit is great.

PR #1219, opened 2018-05-12
Fix AppVeyor and Travis

redmanmale:
* build only master and develop
* update badge link
* disable updater test (it doesn't work)

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1219)


GrayHatter:
Please no... if you're going to comment out large blocks of code, first be VERY sure you actually want to do that. (You don't in this case), then use `#if 0` ... `#endif` via the preprocesser, that way you're not touching every line in the git blame. 

Also, if you want to fix travis, instead of disabling tests, you could create a fake file that you could test against, (probably the best way). Or in this case, add an extra travis job that runs the tests, and allows failure, and then only run the tests in that job.

redmanmale:
If someone fix Travis settings I can just drop the last commit.

redmanmale:
@robinlinden 
Can you take a look?

redmanmale:
It's been a while... Can we merge this?

GrayHatter:
I'll LGTM this, but again, I **really don't** like this. We should create a build job that's allowed to fail, instead of just disabling tests. 

Right now, tests pretends to run, and reports no errors. Thus when anyone looks to see if the test errors, they'll believe that there's no problem.

:lgtm: 

---

Reviewed 2 of 2 files at r1, 1 of 1 files at r2.
Review status: 3 files need 1 more reviewer, missing Admin LGTM at current revision. (waiting on @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, @undefined, and @undefined)

---



*Comments from [Reviewable](https://reviewable.io/reviews/utox/utox/1219#-:-LEzhlpmDVb8PGKA0nKa:bkcdp6w)*

redmanmale:
What will we do with this?

I've taken a quick look at the Travis stages but they are badly compatible with matrix thing.

GrayHatter:
it's stale and @endoffile78 is working on adding new travis builds

PR #1222, opened 2018-05-18
Document the static builds are outdated.

fling-:
Resolves https://github.com/uTox/uTox/issues/1220

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1222)


GrayHatter:
I'm torn on this, it's a trivial merge, but I don't see a real benefit, I'd rather drop the lines if we're not going to maintain build.tox.chat, or fix build.tox.chat if we are

@robinlinden

fling-:
The older binary is still good if you can't build one.

robinlinden:
iirc, they're currently broken because of the git submodules not being included in the source Jenkins provides. I'd like to just fix them rather than dropping them.

GrayHatter:
good luck @robinlinden

robinlinden:
I'm not going to fix this. :(

GrayHatter:
Branch conflict, with different solution than this PR

PR #1264, opened 2018-07-20
Implement "Paste as quote" feature

### LINKED ISSUE #1218: Idea: "paste as quote/greentext"
It so frequently happens, that we are copypasting parts of text from articles or what have you, and we have to prepend every line with a '>' to make it a quote. A context menu item "paste as quote" or "paste as greentext" would be a useful feature.

Thoughts?

avoidr:
- Added entry to context menu
- Bound function to ctrl+shift+v
- Added translations for de.h and pl.h

Closes #1218


avoidr:
I need help with the platform-specific stuff.

PR #1273, opened 2018-08-23
Fix FPS value input field

avoidr:
* snprintf() takes the /size of the buffer/ as argument
* snprintf() returns "the number of bytes that would be written to `s' had `n' been sufficiently large"


avoidr:
Done.

PR #1276, opened 2018-09-03
Unicode documentation and cleanup

avoidr:
I want to review every Unicode-related function.

I'll leave this much for now, to see if someone disagrees so far.


avoidr:
70f6ef54cbf48a69916620dc706d316b3e02ad8b triggers the warning in `utf8_len_read()` about invalid UTF-8 characters.

avoidr:
The problem seems to be that the warning is simply wrong. That code is not unreachable, and it is not bad, that it is not unreachable. It just means that it's not a valid UTF-8 character, but it also does not have to be.

avoidr:
___
*[src/text.c, line 88 at r2](https://reviewable.io/reviews/utox/utox/1276#-LO29xV7-lGKXO31qiSc:-LO2XToJ5A2GvBh8Lp5G:b-gpb80j) ([raw file](https://github.com/utox/utox/blob/55b0222e30e7a0974ca1ca3ced229646b264cacd/src/text.c#L88)):*
<details><summary><i>Previously, GrayHatter (Gregory Mullen) wrote…</i></summary><blockquote>

why did you delete this cases?
</blockquote></details>

Maximum UTF-8 character size is 4 bytes.
Unicode 11.0.0, chapter 2.5 Encoding Forms: "In UTF-8, a character may be expressed with one, two, three, or four bytes"

Issue #1234, opened 2018-06-05
Link-time error at FreeBSD


mazocomp:
### Overview

1. uTox version: latest commit at master
2. Operating system: FreeBSD 11.2-RC1
3. Desktop Environment/Window Manager (Linux/BSD only): CWM

### Issue

Write your issue here.

```
[100%] Linking C executable utox
/usr/bin/ld: undefined reference to symbol `__cxa_demangle@@CXXABI_1.3' (try adding -lcxxrt)
//lib/libcxxrt.so.1: could not read symbols: Bad value
cc: error: linker command failed with exit code 1 (use -v to see invocation)
--- utox ---
*** [utox] Error code 1

make[2]: stopped in /tmp/uTox/build
1 error

make[2]: stopped in /tmp/uTox/build
--- CMakeFiles/utox.dir/all ---
*** [CMakeFiles/utox.dir/all] Error code 2

make[1]: stopped in /tmp/uTox/build
1 error

make[1]: stopped in /tmp/uTox/build
*** [all] Error code 2

make: stopped in /tmp/uTox/build
1 error

make: stopped in /tmp/uTox/build
mazocomp@mazocomp:/tmp/uTox/build %
```

BTW, I should read porting documentation at FreeBSD, net-im/uTox is outdated :/

mazocomp:
After adding -lcxxrt:
```
mazocomp@mazocomp:/tmp/uTox/build % ./utox 
==73692==Sanitizer CHECK failed: /usr/src/contrib/compiler-rt/lib/sanitizer_common/sanitizer_procmaps_freebsd.cc:55 ((Err)) == ((0)) (-1, 0)
mazocomp@mazocomp:/tmp/uTox/build %
```

mazocomp:
Forget what I said, it's no longer true :)

PR #1281, opened 2018-09-10
Some fixes for *BSD systems.

mazocomp:
Depend on find_package() function, do not manually pass -I and -L flags to both compiler and linker.
Allow to build with both D-Bus and filteraudio.
Allow to build and run tests.
Polish build documentation, mention NetBSD and DragonFlyBSD.
Do not set CMAKE_OSX_DEPLOYMENT_TARGET on non-OSX systems.
Fix link-time error at FreeBSD.
Ignore *.orig files, they are generated by patch(1) utility as backup files.
Remove unused OPENBSD variable from CMakeLists.txt.
Do not print "Enable DBus" and "Enable Tests" at Windows and do not print "WIndows Legacy" at not Windows systems.
Enable tests by default.
Fix DragonFlyBSD build.
Make uTox more portable!

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1281)


docs/BUILD.md
```diff
@@ -127,36 +121,31 @@ cd uTox/
...
+
 Now compile uTox:
 
 ```bash
 git clone --recursive git://github.com/uTox/uTox.git
 cd uTox/
 mkdir build
 cd build
-cmake ..
+env CC=gcc cmake ..
```
GrayHatter: I don't want to specify that you need to use gcc here. It should work with any compiler. I know per the IRC conversation that it worked with GCC, but unless you can tell me WHY it only works with GCC. I don't want this in the docs.

CMakeLists.txt
```diff
@@ -303,6 +301,9 @@ elseif(APPLE)
     add_definitions("-x objective-c")
     add_subdirectory(src/cocoa)
 elseif(UNIX)
+    find_package(X11 REQUIRED)
```
GrayHatter: why is this here?

mazocomp: @GrayHatter Because otherwise I have to write the same thing into src/ui/CMakeLists.txt, src/av/CMakeLists.txt, src/layout/CMakeLists.txt and so on in order to pass -I/usr/X11R6/include to compiler in OpenBSD and -I/usr/X11R7/include to compiler in NetBSD, think of this as a local and global scopes, cmake is not that bad as you describe.

docs/BUILD.md
mazocomp: It builds with Clang too, but: https://github.com/uTox/uTox/issues/1234

mazocomp:
I am super porter!

PR #1288, opened 2018-09-29
Styles and fixes

avoidr:
- Fixed two memory leaks
- Checked return values of all `calloc`s in `messages.c`
- Reduced nesting

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1288)

### DIFF
--- a/src/messages.c
+++ b/src/messages.c
@@ -210,39 +210,39 @@ static bool msg_add_day_notice(MESSAGES *m, time_t last, time_t next) {
     ltime_day  = msg_time->tm_mday;
     msg_time   = localtime(&next);
 
-    /* TODO invert logic */
-    if (ltime_year < msg_time->tm_year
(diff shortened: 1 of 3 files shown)


src/messages.c
```diff
@@ -239,25 +251,30 @@ uint32_t message_add_group(MESSAGES *m, MSG_HEADER *msg) {
     return message_add(m, msg);
 }
 
+/* TODO This function and message_add_type_action() are essentially pasta. */
 uint32_t message_add_type_text(MESSAGES *m, bool auth, const char *msgtxt, uint16_t length, bool log, bool send) {
+    FRIEND *f = get_friend(m->id);
+    if (!f) {
+        LOG_DEBUG("Messages", "Could not get friend with id: %u", m->id);
```
GrayHatter: this should never happen, use LOG_ERR when that's the case

PR #1292, opened 2018-10-03
Push to talk works at DragonFlyBSD and FreeBSD

mazocomp:
OpenBSD and NetBSD suck :)

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1292)

### DIFF
--- a/src/xlib/main.c
+++ b/src/xlib/main.c
@@ -94,6 +94,11 @@ void init_ptt(void) {
 
 #ifdef __linux__
 #include <linux/input.h>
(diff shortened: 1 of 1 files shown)


endoffile78:
Linux doesn't have that header.

mazocomp:
@endoffile78 huh, *BSD systems have a stupid habit to rename headers taken from Linux. OpenBSD and NetBSD don't have evdev, but somehow Mumble supports push-to-talk there.

PR #1293, opened 2018-10-03
Simplify save loading code

robinlinden:
This makes the saving code obviously never return -2 if we see a tox.save or any variant of it. Before this PR, it (at least) looks like you can reach the end of this block https://github.com/uTox/uTox/blob/6a995f2c576bc24c79d3dd245b49bde49ef085a7/src/tox.c#L294 and return -2 even if we saw a tox.save on disk. It looks like it's impossible, but probably better to make it obviously impossible.

This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1293)


GrayHatter:
:LGTM:

PR #1302, opened 2018-10-13
Fix large file transfers on Windows

### LINKED ISSUE #1298: Windows: send file size bug
### Overview

1. uTox version: 0.16 -> master
2. Operating system: windows

### Issue

If file size is larger than 0x7FFFFFFF conversion between size_t to uint64_t results to this (t) bug - file size became negative number.
i suggest to change size_t size = ftello(file); -> uint64_t size = ftello64(file); in file_transfers.c

redmanmale:
as @Vityacv proposed. Fixes #1298.

Also add parentheses to file size format: ```filename [1 GiB]```

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1302)


avoidr:
`file_pos()` should return `off_t`, not `size_t`.

Otherwise, looks good.
Also good idea with the brackets.

redmanmale:
It doesn't work if I change it to ```off_t```.

avoidr:
Works for me on musl. What's the issue?

redmanmale:
https://github.com/redmanmale/uTox/blob/fix-get-file-size/src/file_transfers.c#L1119

If this ```size``` is ```off_s``` everything stops working. Wrong value after cast somehow.

avoidr:
I don't see why that would happen. What's the size of the file? Have you tested on POSIX, too?

redmanmale:
Commited code works both on Windows and Ubuntu.

After that I'm testing only on Windows without success.
Maybe I'm doing something wrong? Could you publish your changes to some branch and I'll test it on Windows myself?

avoidr:
This is all I meant to do: https://github.com/avoidr/uTox/commit/5d8fe5f7cf181101b12e773ba3edca65363a2181

Note that I did not test myself, but I do not see how that would break anything.

redmanmale:
Just checked, your version does not work.
I suspect it was caused by this cast https://github.com/avoidr/uTox/blob/5d8fe5f7cf181101b12e773ba3edca65363a2181/src/windows/filesys.c#L95

avoidr:
I really have no explanation for that. I see nothing. How do you test? Is the failure that it behaves just like without any patches?

It seems to me that off_t is something too small, even though it says "Represents file-offset value". _ftelli64() returns __int64, which coincides with what ftello() returns (currently).
Seeing how tox_file_send() takes a uint64_t for the file size, I think returning a uint64_t would be the next correct thing. I don't think size_t is really correct semantically.
Or, how about an #ifdef for the Windows-specific file_pos() signature?

redmanmale:
I just run uTox with and without this change.
And there's a correct file size in UI without it and the wrong one with it.

redmanmale:
Works fine with ```uint64_t```

redmanmale:
It's different approach now.

redmanmale:
Bump.

redmanmale:
Rebased.

redmanmale:
Rebased.

src/text.c
```diff
@@ -22,13 +22,13 @@ int sprint_humanread_bytes(char *dest, unsigned int size, uint64_t bytes) {
 
     size_t r;
 
-    r = snprintf((char *)dest, size, "%u", (uint32_t)bytes);
+    r = snprintf((char *)dest, size, "[%u", bytes);
```
avoidr, line 25: Please use `PRIu64` format specifier from `inttypes.h` here.

avoidr, line 25: Please use `PRIu64` format specifier from `inttypes.h` here.

avoidr:
Please rebase when you're done. That's because I don't want the forest that we once had. We have a linear git log now, except when a branch consists of multiple commits, then we merge --no-ff to "group" those changes. Easier to skip them when skimming the log.

I'd rebase that stuff myself, but then signatures get lost.

src/text.c
redmanmale: Done.

avoidr:
Exactly like that, thanks.

avoidr:
Rebase again, then I can finally merge.

avoidr:
Merged, thanks!

PR #1307, opened 2018-10-27
[Windows] Wait until window is inited before coping ToxID from Toxcore

### LINKED ISSUE #592: Tox ID isn't displayed
Tox ID field looks like this before selection:

after selection:

But ID could not be copied either via CTRL-C or button COPY.

Win 7, 8.1 x64, uTox v.0.11.1

redmanmale:
Fixes #592.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1307)

### DIFF
--- a/src/tox.c
+++ b/src/tox.c
@@ -442,10 +442,11 @@ static int init_toxcore(Tox **tox) {
  *
  * Accepts and returns nothing.
  */
-void toxcore_thread(void *UNUSED(args)) {
+void toxcore_thread(void *args) {
     ToxAV *av               = NULL;
     bool   reconfig         = 1;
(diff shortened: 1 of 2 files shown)


GrayHatter:
What's the goal of this?

I hate the idea of blocking on toxcore thread waiting for the main thread...

redmanmale:
The goal is to avoid this thing from happening.
In that case user is unable to see or copy his ToxID.
error

GrayHatter:
how about preloading that text entry field with "Loading ToxID..." instead?

redmanmale:
I'll try. Just didn't think about it.

GrayHatter:
@redmanmale bump

PR #1317, opened 2018-11-19
Add support for loading groupchats

endoffile78:
~Currently it loads the groups but when you try to send a message it fails.~

Toxcore doesn't allow messages to be sent to a groupchat with only one person in it, this just needs to be tested with a groupchat with more than one person in it.

This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1317)


robinlinden:
I've tested the PR a bit with a few different groups and it seems to work well.

PR #1320, opened 2019-01-21
Use evdev only if it's present.

mazocomp:
Otherwise fallback to Xlib at every X11 platform, but support push-to-talk everywhere!

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1320)

### DIFF
--- a/src/xlib/main.c
+++ b/src/xlib/main.c
@@ -98,8 +98,8 @@ void init_ptt(void) {
 #include <dev/misc/evdev/input.h>
 #endif
 
+static bool check_ptt(void) {
 #if defined(__linux__) || defined(__DragonFly__) || defined(__FreeBSD__)
-static bool linux_check_ptt(void) {
     /* First, we try for direct access to the keyboard. */
     int ptt_key = KEY_LEFTCTRL; // TODO allow user to change this...
     if (ptt_keyboard_handle) {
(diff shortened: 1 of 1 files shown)


GrayHatter:
why?

mazocomp:
@GrayHatter why was evdev even used if existing Xlib code just works?

GrayHatter:
because evdev works on both wayland and x11, and that code was written before I knew/was willing to use libinput.

GrayHatter:
And evdev is is A LOT faster than xlib.

mazocomp:
@GrayHatter damn you, if you want to support Wayland, why do you support it in Xlib module? Also evdev is Linux-specific thing which both FreeBSD and DragonFlyBSD borrow.

GrayHatter:
Because I wasn't ready to write all the xlib code when I was implementing PTT, so that was the best location at the time.

mazocomp:
@GrayHatter the best thing I've done today was removing code to not use shit called evdev at all. But then you made me do the worst thing: use evdev only at Linux, FreeBSD and DragonFlyBSD, at other systems use only Xlib.

mazocomp:
@GrayHatter oh, I see the quality of uTox doesn't matter to you at all.

GrayHatter:
If you'd actually work just a little to understand why the current code base actually exists before making PRs I'd be happy to accept them. But I have no interest in wasting time on low effort commits that fix nothing. Especially considering the arrogance from conversation the doesn't match the quality of the patch. If that means it lowers the quality in your opinion. Then nothing of value is lost.

GrayHatter:
I just remembered another reason we us edev. X allows another app to take exclusive control of the keyboard. this way, even if there's an app that has exclusive control of the keyboard, ptt in uTox will still work.

PR #1325, opened 2019-02-13
Fix for buffer overflow on app startup

ColumPaget:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1325)


ColumPaget:
What the hell? Something seems to have gone very wrong with this pull request? Where did all this stuff come from? I'm only changing one file!

PR #1326, opened 2019-02-13
Fix for buffer overflow on app startup

ColumPaget:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1326)


avoidr:
The last `if` should have been inverted. My bad.

PR #1328, opened 2019-02-21
Make push-to-talk work at every X11 platform.

mazocomp:
DragonFly BSD changed location for evdev input.h file, FreeBSD has different location from the beginning. Note that DragonFly BSD is compiled without evdev support by default, but it still has that header.

### DIFF
--- a/src/xlib/main.c
+++ b/src/xlib/main.c
@@ -92,14 +92,16 @@ void init_ptt(void) {
 
 
 
-#ifdef __linux__
+#if defined(__linux__) || defined(__DragonFly__)
 #include <linux/input.h>
-#elif defined(__DragonFly__) || defined(__FreeBSD__)
-#include <dev/misc/evdev/input.h>
+#define HAVE_EVDEV
+#elif defined(__FreeBSD__)
+#include <dev/evdev/input.h>
+#define HAVE_EVDEV
 #endif
 
-#if defined(__linux__) || defined(__DragonFly__) || defined(__FreeBSD__)
 static bool linux_check_ptt(void) {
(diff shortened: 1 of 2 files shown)


src/xlib/main.c
```diff
@@ -92,14 +92,16 @@ void init_ptt(void) {
...
-#include <dev/misc/evdev/input.h>
+#define HAVE_EVDEV
+#elif defined(__FreeBSD__)
+#include <dev/evdev/input.h>
+#define HAVE_EVDEV
 #endif
 
-#if defined(__linux__) || defined(__DragonFly__) || defined(__FreeBSD__)
 static bool linux_check_ptt(void) {
+#ifdef HAVE_EVDEV
```
GrayHatter: I don't like inline ifdefs

Plus, it's not linux_check_ptt if it's running on bsd

mazocomp: right, I forgot about renaming the function

src/xlib/main.c
```diff
@@ -116,12 +119,26 @@ static bool linux_check_ptt(void) {
...
+
+    ppt_keyboard_handle_unknown = true;
+    return false;
+}
+
+static bool check_ptt(void) {
+    /* First, we try for direct access to the keyboard. */
+    if (ppt_keyboard_handle())
+        return true;
+    else if (!ppt_keyboard_handle_unknown)
```
GrayHatter: 1TBS

Needs { }

mazocomp: ok, done

mazocomp:
@GrayHatter please remind me the link to branch, I can't find it.

PR #1334, opened 2019-03-05
Use cmake timestamp function

bmwiedemann:
Use cmake timestamp function:
This makes it simpler, work on more platforms
and allows for reproducible builds.
See https://reproducible-builds.org/ for why this is good.

Without this patch, I got such diffs
```diff
/usr/share/man/man1/utox.1.gz
@@ -1,4 +1,4 @@
-.TH UTOX "1" "March 2019" "µTox 0.17.0"
+.TH UTOX "1" "April 2034" "µTox 0.17.0"
```

This PR was done while working on reproducible builds for openSUSE.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1334)


GrayHatter:
Cheers mate, thanks!!

PR #1339, opened 2019-04-11
qr_setup() fixes

avoidr:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1339)


avoidr:
>alternitivlly, maybe tox_uri should exist on the stack?

Yeah, that's best, especially since its size is a constant.

>memory alloc fails are usually fatal. Do you want to make this fatal, or should we reconsider that SOP?

This is what I was thinking: It feels intuitively wrong to me to exit the program because of what is essentially only a convenience feature. QR is not crucial to the main goal of utox. Therefore I don't see it as a reason to panic everything.
Also, it would just look like utox is unstable, and nobody likes that.

avoidr:
I'm done.

avoidr:
>optionally prefer sizeof("tox:") + ... because that'll include the null term.

I did it intentionally like this because I dislike VLA and I like being explicit. Anyone (even me) can immediately see that it's correct.

PR #1344, opened 2019-04-18
Fix travis

GrayHatter:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1344)


robinlinden:
I don't recall what this PR was fixing, but Travis seems to pass at HEAD of develop. @GrayHatter do you remember what was going on with Travis back here?

avoidr:
And here, delete branch?

Issue #1279, opened 2018-09-08
crash pn windows id > 31


zoff99:
https://github.com/uTox/uTox/blob/55b0222e30e7a0974ca1ca3ced229646b264cacd/src/xlib/video.c#L22

on ubuntu 18.04 i get values over 31. how are these window IDs generated?

zoff99:
window id seems to be the friendnumber which can be up to UINT32_MAX
so i fear this needs a different logic

zoff99:
https://github.com/uTox/uTox/blob/55b0222e30e7a0974ca1ca3ced229646b264cacd/src/windows/main.h#L54-L55
fix here too. at least here you can have 128 friends before it crashes

zoff99:
hello?

zoff99:
yes, have more than 32 friends and start video call with the friend you added last.
then uTox will access the array out of bounds without checking. and corrupt its memory later at some random position.

a) the MAX_VID_WINDOWS should at least be 128 (like in the windows version)
b) check for array out of bounds access and give an error, instead of corrupting memory

zoff99:
https://github.com/uTox/uTox/blob/55b0222e30e7a0974ca1ca3ced229646b264cacd/src/xlib/video.c#L48

here check if id < MAX_VID_WINDOWS

bascially check at every access of "video_win" array if its would be accessed out of bounds

zoff99:
in the end this needs to be totally rewritten, it cant work properly with a fixed size array

GrayHatter:
yeah, uTox stores system window references in a static complied defined array. Someone needs to write a realloc helper that'll dynamically allocate a buffer to store all possible system window references. I think I wrote the skel for that helper, but I never plumbed it together to actually do the dynamic array thing.

A short term fix could be just increasing that array size to 2^8 or 2^7 until the code is written.

avoidr:
@zoff99, can you please see if it at least prevents the crashes?

PR #1350, opened 2019-04-21
Prevent crash when Window ID is too large

avoidr:
Bound-check video_win[]

See #1279

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1350)


avoidr:
I am aware that it doesn't solve the root problem, but I am not trying to here. I just don't want it to crash.

PR #1352, opened 2019-04-22
Check return values of allocations

avoidr:
This much for now, for this PR.

Note that in most cases I just made it `FATAL`, but I can't judge if it's right.
Feel free to fix it up yourself, might be easier (branch is on this repo).

`$ ack --color '[mce]alloc\(' | cut -d':' -f1 | uniq -c | sort -n`

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1352)

### DIFF
--- a/src/devices.c
+++ b/src/devices.c
@@ -18,7 +18,6 @@ static bool realloc_devices_list(uint16_t new_size) {
     }
 
     UTOX_DEVICE *tmp = realloc(devices, sizeof(UTOX_DEVICE) * new_size);
-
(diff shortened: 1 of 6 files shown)


avoidr:
I *did* consider \n there. But then this one check stole focus and takes up 50% of the block. Didn't find that pretty, either. Whatchu think?

avoidr:
But I guess if only the braces are a problem in this case, then I'd like that best, single line and no braces.

PR #1355, opened 2019-05-11
Add border to QR code

### LINKED ISSUE #1351: QR code is hard to scan on dark background/themes
Scanning is practically impossible against dark backgrounds.
Scanning works flawlessly with light themes.

avoidr:
This is necessary for contrast, so that the QR code becomes readable on any background.

Fixes #1351

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1355)

### DIFF
--- a/src/qr.c
+++ b/src/qr.c
@@ -12,6 +12,8 @@
 #include <stdlib.h>
 #include <string.h>
 
+#define QR_BORDER_SIZE 4 /* unit: QR modules */
+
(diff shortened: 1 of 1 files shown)


redmanmale:
Some screenshots would be nice.

avoidr:
This is how it looks now:
image

But the thing is, now it's actually a correct QR code. The border thing is called Quiet Zone and it's a necessary part. And the width is recommended to be 4 modules wide.

PR #1358, opened 2019-05-12
Fix segfault when changing to a video input device that fails to init

avoidr:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1358)

### DIFF
--- a/src/av/video.c
+++ b/src/av/video.c
@@ -28,7 +28,7 @@ static bool     video_active         = false;
 
 static utox_av_video_frame utox_video_frame;
 
-static bool video_device_status = true;
+static bool video_device_status = false;
 
 static vpx_image_t input;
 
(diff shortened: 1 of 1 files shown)


src/av/video.c
```diff
@@ -142,26 +141,31 @@ bool utox_video_change_device(uint16_t device_number) {
...
-        return true;
-    } else {
-        /* Just grab the new frame size */
-        close_video_device(video_device[video_device_current]);
+        goto fail;
     }
+
+    pthread_mutex_unlock(&video_thread_lock);
+    return true;
+fail:
```
GrayHatter: indentation needed

GrayHatter: also, probably needs a better name

avoidr: Done.
Indentation like this?

PR #1359, opened 2019-05-17
Simplfy regexes that match MacOS, Linux, NetBSD and FreeBSD

mazocomp:


robinlinden:
So the account that made this PR is now gone, but I'm not convinced this is correct?
For reference: https://skia.googlesource.com/third_party/sdl/+/refs/heads/master/CMakeLists.txt#84
This is how a lot of other projects detect which platform they're building on.

avoidr:
@goleo108

goleo108:
@robinlinden @avoidr regular expressions are designed to match substrings, having a string "kbgkfafdsfnas" and regexp "faf" will match "faf", having regexp ".*faf" will match "kbgkfaf", regexp "faf.*" will match "fafdsfnas" and regexp ".*faf.*" will match entire string "kbgkfafdsfnas"

goleo108:
just to be clear: there is no need to match entire string. matching only part of the string is enough, pattern "FreeBSD" will match any string containing "FreeBSD", for example, "kFreeBSD 12.0-STABLE some version info", you don't need extra patterns like `"^k?FreeBSD [0-9]+\.[0-9]+-(RELEASE|STABLE|CURRENT) .+$"`

avoidr:
I merged it, thanks @goleo108! Sorry it took so long.

Btw, how do you want to be credited in release notes? Old nick, new nick, something else?

goleo108:
@avoidr may it be like everyone else: point to my current gh account

avoidr:
Sure thing.

PR #1382, opened 2019-10-27
Focus message input field when starting to type

### LINKED ISSUE #1142: Focus message input field when starting to type
There is no way that I see to set focus to the message input field with the keyboard. One has to use the mouse and click on the field.
I propose to set focus on keypress, so you never have to bother and typing just works.

avoidr:
Closes #1142

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1382)

### DIFF
--- a/src/xlib/event.c
+++ b/src/xlib/event.c
@@ -443,7 +443,7 @@ bool doevent(XEvent *event) {
                     break;
                 }
 
-                if (ev->state & 4) {
+                if (ev->state & ControlMask) {
                     if (sym == 'c' || sym == 'C') {
                         if (flist_get_friend()) {
                             clipboard.len = messages_selection(&messages_friend, clipboard.data, sizeof(clipboard.data), 0);
@@ -503,7 +503,7 @@ bool doevent(XEvent *event) {
                     sym = XK_Return;
                 }
 
-                if (sym == XK_Return && (ev->state & 1)) {
+                if (sym == XK_Return && (ev->state & ShiftMask)) {
                     edit_char('\n', 0, 0);
                     break;
                 }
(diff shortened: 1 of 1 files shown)


avoidr:
Not yet.

avoidr:
I will clean the history in another PR/branch before I merge, but here is my result. I think I'm done now.

src/xlib/event.c
```diff
@@ -415,29 +415,56 @@ bool doevent(XEvent *event) {
...
+                break;
+            }
+
+            if (!edit_active()) {
+                if (messages_char(sym)) {
                     redraw();
                     break;
                 }
+
+                if (ev->state & 4) {
```
GrayHatter: can we make this a macro? I assume that `& 4` tests for ctrl?

src/xlib/event.c
```diff
@@ -415,29 +415,56 @@ bool doevent(XEvent *event) {
...
+                /* Focus message input field if ctrl isn't pressed,
+                 * to make sure you can still copy text from the chat log */
+                if (sym != XK_Control_L) {
+                    edit_setfocus(&edit_chat_msg_friend);
+                    edit_char(KEY_END, 1, 0);
+                }
             }
 
             if (edit_active()) {
+                if (sym == XK_Escape) {
```
GrayHatter: what's Esc doing here?

src/xlib/event.c
avoidr: Yeah, actually that is `ControlMask` from Xlib.

src/xlib/event.c
avoidr: It's explained in the commit message, that block is not directly related to focusing, somewhat a separate "feature".

avoidr:
Cleaned up and merged.

PR #1387, opened 2019-11-02
Handle Home & End key in messages panel

avoidr:
Home now scrolls to top    of message log
End  now scrolls to bottom of message log

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1387)


avoidr:
Merged.

PR #1391, opened 2019-11-07
Fix Up key

avoidr:
Now, pressing Up when the cursor is already on the first line makes the cursor go to the beginning of the line.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1391)


GrayHatter:
:LGTM:

PR #1403, opened 2019-12-08
Ontab to search

avoidr:


avoidr:
Rebased.

avoidr:
onshifttab callback doesn't work in group, if the group is empty.

PR #1407, opened 2019-12-15
Remove WinXP support

avoidr:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1407)


avoidr:
Thanks!

avoidr:
LInes deleted: 404
Perfect.

victordesfe:
petition to include a compillable old version that works with xp.
me and other folks (username1565) have been trying unsuccesfully to do that for a while, but now this decision fucks us up even more in the try while you offer little to no help in that regard.
to date it is near to impossible to compile from source any past version.
thanks for your comprehension

avoidr:
Microsoft dropped support for Windows XP. So the right place to complain is at Microsoft.

uTox is software with a strong focus on security. Supporting an obsolete and rotting operating system would send bad signals and put our users at risk, *and Tox users as a whole*. uTox won't be responsible for that. **uTox will never support Windows XP anymore.**

robinlinden:
If you want a version of uTox that compiles for Windows XP, your best bet is to check out https://github.com/uTox/uTox/commit/0bcfead50a2ef3009440bff9a7188cb605b91c44 and build that. It's the commit before the one that removed support for Windows XP.

endoffile78:
@victordesfe If possible please consider upgrading to a new version of Windows. Windows XP is no longer supported by Microsoft and does not receive security update putting yourself at risk to being the victim of exploits found after Microsoft quit supporting XP. If you do not like newer versions of Windows or they don't run well with the hardware you have then consider switching to Linux. Ubunutu or Linux Minut would be a good beginner distro.

PR #1413, opened 2020-01-23
Mark global variable as extern

### LINKED ISSUE #1412: GCC 10 compatibility issue
With the latest GCC 10 (in Fedora Rawhide), uTox fails to build, see extensive log here: 
https://kojipkgs.fedoraproject.org//work/tasks/1215/40891215/build.log

This is related to this change in GCC 10:

GCC now defaults to -fno-common. As a result, global variable accesses are more efficient on various targets. In C, global variables with multiple tentative definitions now result in linker errors. With -fcommon such definitions are silently merged during linking.

https://gcc.gnu.org/gcc-10/changes.html

Default to -fno-common

A common mistake in C is omitting extern when declaring a global variable in a header file. If the header is included by several files it results in multiple definitions of the same variable. In previous GCC versions this error is ignored. GCC 10 defaults to -fno-common, which means a linker error will now be reported. To fix this, use extern in header files when declaring global variables, and ensure each global is defined in exactly one C file. As a workaround, legacy C code can be compiled with -fcommon.

  int x;  // tentative definition - avoid in header files

  extern int y;  // correct declaration in a header file

https://gcc.gnu.org/gcc-10/porting_to.html

eclipseo:
GCC now defaults to -fno-common. As a result, global variable accesses
are more efficient on various targets. In C, global variables with
multiple tentative definitions now result in linker errors.

Fix #1412

Signed-off-by: Robert-André Mauchin <zebob.m@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1413)


avoidr:
Rebased.

avoidr:
Thanks!

PR #1422, opened 2020-03-01
Travis: Linux: disable test_updater

avoidr:
The test now fails due to the utox.io domain being expired.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1422)


avoidr:
Merged.

PR #1426, opened 2020-03-09
Don't pass 2 of every flags in CFLAGS

robinlinden:
The variable CMAKE_C_FLAGS is initialised from CFLAGS by CMake, see:
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1426)


avoidr:
Thanks!

PR #1427, opened 2020-03-10
WIP: Use FetchContent to pull in third-party dependencies

robinlinden:
Not sure if this is something we want to do, but I wanted to have a look at how our CMake would look like without building random libraries into main uTox.

Per https://cmake.org/cmake/help/latest/module/FetchContent.html you can e.g. set `FETCHCONTENT_SOURCE_DIR_libname` to an alternative source dir to stop CMake from downloading anything. Another improvement might be to add a variable to use `Findlibname` instead of the FetchContent-stuff.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1427)


robinlinden:
Closing. Will reopen if this is something we want to do at some point in the future.

PR #1428, opened 2020-03-13
Static libs

avoidr:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1428)


avoidr:
This batch is done now.

PR #1434, opened 2020-04-01
Remove auto-updater

avoidr:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1434)


avoidr:
Rebased.

avoidr:
Rebased.

avoidr:
Squashed.

GrayHatter:
I'm not sure I would have approved this change...

Why was deleting everything better than trying to fix it, or comment it out?

avoidr:
Windows is unmaintained and there are no updates to uTox, anyway. Nobody would have fixed it for a long time (if ever) and we don't even own that domain anymore. And I don't want to kid anybody. It's just one more broken promise that disappoints and makes a bad impression. And I don't want to see uTox in such a bad state, either.

For the little activity that uTox is seeing, @robinlinden may build binaries for releases. Speaking of releases, I want to do them more often and more regularly, too.

PR #1448, opened 2020-04-16
Fix Travis

robinlinden:
I broke it in the last PR, and when trying to fix it it couldn't download libshell32 from cmdline.org and it seems to build without that download anyway?

We should re-enable the pre-merge Travis check. :D

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1448)


redmanmale:
Are these switches on in uTox project settings on Travis?
image

robinlinden:
Yeah, it's all enabled and it builds on every PR, we just don't get the status here. :(
https://travis-ci.org/github/uTox/uTox

Travis is an optional check on PRs though. I'll try making it mandatory and see if that helps it run.

PR #1452, opened 2020-04-20
Rename some flist functions

avoidr:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1452)


GrayHatter:
@avoidr selected != current. you have have selected one friend, but a different one could be current

e.g. ```
mouse down selects
utox can do something else (validate state)
mouse up changes the current from old -> new
```

I don't think you should change it, I'm just being pedantic.

PR #1456, opened 2020-04-25
CHANGELOG.md: redirect to release_notes/

avoidr:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1456)


redmanmale:
Why?

avoidr:
Because our release notes are actually user-friendly and user-relevant. And our linear/grouped git log is now actually useful for developers, too.

And our changelog… is just a dump of whatever happened on github. And the script that generates it isn't even deterministic, last I remember.

redmanmale:
Maybe you're right.

But I've got some thoughts on the release notes format.
Commit Id (or multiple) at the end of the line doesn't look very user-friendly and it's not easy to use them since you have to copy-paste it by hand.
I think links to fixed issues and PRs in the Fixes and Features sections respectively look much more nice and handy.

avoidr:
But what's the real benefit of linking to github issues and PRs? All the relevant info and conclusions should be included in the git log. Otherwise it's a documentation problem. If git log has the info, it should be enough to just give the relevant commit IDs.

I put only the commit hashes at the end, because they are forge-independent, and forges (or at least the one we're currently on, github), automatically make links of them to commits, anyway, and that independence goes away if we made the link explicit. And if you're reading those release notes locally, likely those URLs to a website are of no use to you and you don't need that clutter that comes with them. That would make it actually unusable for the people for whom that info is primarily for.
Basically, by not making the links explicit and absolute, they are always relative, and in my opinion appropriate for the UI they're viewed on.

By the way, my current process of creating release notes is copying relevant lines from git log (`--format=format:"%s (@%an) (%H) %H" --graph <tag>..`¹), perhaps adjusting the title to add context for users. That's trivial and easy enough to do mechanically — and redo, consistently and easily verifiably — even on the last day before release after months of slacking around. Making that process more tedious again would get us back to square one (no releases for a long time).

¹ Note that the trailing `%H` is just for easy selecting the commit hash with a double-click, while creating the notes.

PR #1458, opened 2020-04-29
Refactoring save and load settings

### LINKED ISSUE #1457: Sometimes uTox config cuts off
### Overview

1. uTox version: 0.17.2
2. Operating system: Win 10

### Steps to reproduce

* get fresh uTox instance
* run it in portable mode
* close after sometime (doesn't really matter, just give it enough time to init properly)

In the Tox folder will be three files:
* tox_save.tox - OK
* utox_save.in~ - temp save file created by minIni
* utox_save.ini - cuts off like this:

```
[general]
save_version=3
utox_last_version=4354
send_version=false
update_to_develop=false
```

Sometimes there are fewer (1-2) or more lines (10-15).
Next time (run and close) there could be a little more lines or no file at all.

But once config would be successfully saved there won't be any further issues.

redmanmale:
* drop old and deprecated settings binary format
* drop `UTOX_SAVE` struct because its only purpose was hold data from `SETTINGS` struct and dump it to the disk
* drop unused function `config_osdefaults`
* all fields in the `SETTINGS` struct renamed to match fields from `UTOX_SAVE` because they are loaded from the file with `NAMEOF` macro and we don't want return to the tons of the consts strings
* drop unused parameters (`auto_update`, ...) and add missing (`block_friend_requests`)
* save settings after failed load attempt (uTox first run)
* save settings right after begining of the shutdown process (fixes #1457)
* enforce codestyle

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1458)

### DIFF
--- a/src/av/audio.c
+++ b/src/av/audio.c
@@ -703,7 +703,7 @@ void utox_audio_thread(void *args) {
                     break;
                 }
                 case UTOXAUDIO_PLAY_RINGTONE: {
-                    if (settings.ringtone_enabled && self.status != USER_STATUS_DO_NOT_DISTURB) {
+                    if (settings.audible_notifications_enabled && self.status != USER_STATUS_DO_NOT_DISTURB) {
                         LOG_INFO("uTox Audio", "Going to start ringtone!" );
 
                         audio_out_device_open();
@@ -727,7 +727,7 @@ void utox_audio_thread(void *args) {
                     break;
                 }
                 case UTOXAUDIO_PLAY_NOTIFICATION: {
-                    if (settings.ringtone_enabled && self.status == USER_STATUS_AVAILABLE) {
+                    if (settings.audible_notifications_enabled && self.status == USER_STATUS_AVAILABLE) {
                         LOG_INFO("uTox Audio", "Going to start notification tone!" );
 
                         if (close_device_time <= time(NULL)) {
@@ -783,7 +783,7 @@ void utox_audio_thread(void *args) {
             }
         }
 
-        settings.audiofilter_enabled = filter_audio_check();
+        settings.audio_filtering_enabled = filter_audio_check();
 
         bool sleep = true;
 
@@ -811,7 +811,7 @@ void utox_audio_thread(void *args) {
 
             #ifdef AUDIO_FILTERING
             #ifdef ALC_LOOPBACK_CAPTURE_SAMPLES
-            if (f_a && settings.audiofilter_enabled) {
+            if (f_a && settings.audio_filtering_enabled) {
                 alcGetIntegerv(audio_out_device, ALC_LOOPBACK_CAPTURE_SAMPLES, sizeof(samples), &samples);
                 if (samples >= perframe) {
                     int16_t buffer[perframe];
--- a/src/main.c
+++ b/src/main.c
@@ -262,13 +262,13 @@ void utox_init(void) {
         settings.debug_file = stdout;
     }
 
-    UTOX_SAVE *save = config_load();
-    free(save);
+    config_load();
 
     /* Called by the native main for every platform after loading utox setting,
      * before showing/drawing any windows. */
     if (settings.utox_last_version != settings.last_version) {
         settings.show_splash = true;
+        settings.utox_last_version = settings.last_version;
     }
 
     // We likely want to start this on every system.
--- a/src/main.c
+++ b/src/main.c
@@ -267,7 +267,7 @@ void utox_init(void) {
 
     /* Called by the native main for every platform after loading utox setting,
      * before showing/drawing any windows. */
-    if (settings.curr_version != settings.last_version) {
+    if (settings.utox_last_version != settings.last_version) {
         settings.show_splash = true;
     }
 
--- a/src/settings.c
+++ b/src/settings.c
@@ -16,7 +16,8 @@
 #include "native/filesys.h"
 #include "native/keyboard.h"
 
-#include "main.h" // UTOX_VERSION_NUMBER, MAIN_HEIGHT, MAIN_WIDTH, all save things..
+// UTOX_VERSION_NUMBER, MAIN_HEIGHT, MAIN_WIDTH, all save things..
+#include "main.h"
 
 #include <stdlib.h>
 #include <string.h>
@@ -25,7 +26,8 @@
 #define MATCH(x, y) (strcasecmp(x, y) == 0)
 #define BOOL_TO_STR(b) b ? "true" : "false"
 #define STR_TO_BOOL(s) (strcasecmp(s, "true") == 0)
-#define NAMEOF(s) strchr((const char *)(#s), '>') == NULL ? #s : (strchr((const char *)(#s), '>') + 1)
+#define NAMEOF(s) strchr((const char *)(#s), '>') == NULL \
+    ? #s : (strchr((const char *)(#s), '>') + 1)
 
 static const char *config_file_name = "utox_save.ini";
 
@@ -60,9 +62,8 @@ SETTINGS settings = {
 
     // Low level settings (network, profile, portable-mode)
     .disableudp     = false,
-    .enableipv6    = true,
-
-    .proxyenable      = false,
+    .enableipv6     = true,
+    .proxyenable    = false,
     .force_proxy    = false,
     .proxy_port     = 0,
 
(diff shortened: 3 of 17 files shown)


avoidr:
I think with d3098441f0349bf58eeb50bea1af55af4c783241 you can also remove `DEFAULT_{WIDTH,HEIGHT}` and perhaps even `DEFAULT_SCALE`.

redmanmale:
> 
> 
> I think with [d309844](https://github.com/uTox/uTox/commit/d3098441f0349bf58eeb50bea1af55af4c783241) you can also remove `DEFAULT_{WIDTH,HEIGHT}` and perhaps even `DEFAULT_SCALE`.

Done.

src/settings.h
```diff
@@ -57,90 +50,45 @@ typedef struct utox_settings {
...
     bool    status_notifications;
     uint8_t group_notifications;
 
+    uint16_t audio_device_out;
+    uint16_t audio_device_in;
+    uint8_t  video_fps;
+
     LOG_LVL verbose;
-    FILE *debug_file;
+    FILE *  debug_file;
```
avoidr, line 66: Why this? Looked fine before.

src/settings.h
```diff
@@ -57,90 +50,45 @@ typedef struct utox_settings {
...
-    uint8_t force_proxy;
-    uint8_t use_long_time_msg;
-
-    uint8_t  unused[51];
-    uint8_t  proxy_ip[];
-} UTOX_SAVE;
-
 /*
- * Loads the config file and returns a settings struct
+ * Loads setting from the disk
```
avoidr, line 85: I would personally just say "from disk". And "settings".

src/settings.h
```diff
@@ -57,90 +50,45 @@ typedef struct utox_settings {
...
  */
-UTOX_SAVE *config_load(void);
+void config_load(void);
 
 /*
- * Writes save_in to the disk
+ * Writes settings to the disk
```
avoidr, line 90: Same here, "to disk".

avoidr:
I see you have a lot of style fixes here: 48bb4614190ede98639304f1994959336fae40d9
Can you please keep them in a separate commit?

redmanmale:
> Can you please keep them in a separate commit?

I almost completely rewrote settings .h and .c with new codestyle.
And to keep this two changes separated I would have to do it all again.

So if you don't insist I'd like not do it.

src/settings.h
redmanmale: For vertical alignment (as it was before).

src/settings.h
redmanmale: Done.

src/settings.h
redmanmale: Done.

src/settings.h
avoidr: Ok, right, somehow I saw there was a line between them.

avoidr:
If you don't mind, I'll split the style changes myself on my own branch, then you can take them if you like it. I'll do that and review later today.

redmanmale:
Ohh, now I feel a little embarrassed.
And yeah, it would be great, thanks mate!

avoidr:
No, it's not like that! It's just that I don't like bothering people with stuff that I want. I would feel awkward myself :P So I prefer to help out a bit rather than make you do annoying things after all this effort you already did.

avoidr:
I've split the commit in my branch: [avoidr/uTox:save-file-refactoring](https://github.com/avoidr/uTox/tree/save-file-refactoring)
I've made the new commit have the same timestamp as the original one. Take a look.

src/settings.c
```diff
@@ -283,293 +288,231 @@ static int config_parser(const char* section, const char* key, const char* value
...
     }
 
     free(config_path);
 
-    return save;
+    return true;
 }
 
-static bool utox_save_config(UTOX_SAVE *config) {
+static bool utox_save_config() {
```
avoidr, line 310: Empty parameter list needs to be `void`.

avoidr:
What do you think about creating macros for the `write_config_value_*` functions?

redmanmale:
> What do you think about creating macros for the write_config_value_* functions?

I'm not sure I understand how to write it.
These funcs have different arg types and differs inside.

avoidr:
Yeah, I meant one macro for each function. I believe the lines will be halved and much easier to read and verify with such macros.

src/av/audio.c
```diff
@@ -703,7 +703,7 @@ void utox_audio_thread(void *args) {
                     break;
                 }
                 case UTOXAUDIO_PLAY_RINGTONE: {
-                    if (settings.ringtone_enabled && self.status != USER_STATUS_DO_NOT_DISTURB) {
```
GrayHatter, line 706: This behavior is intentional. We wanted to support being able to have audible alerts for notification or ringtone. This change changes that.

redmanmale: But it's just rename...

src/main.c
```diff
@@ -261,13 +261,13 @@ void utox_init(void) {
         settings.debug_file = stdout;
     }
 
-    UTOX_SAVE *save = config_load();
-    free(save);
+    config_load();
```
avoidr, line 264: Why is `settings` accessed before config was loaded?

tests/mock/mock_settings.c
```diff
@@ -60,32 +60,13 @@ SETTINGS settings = {
 /*
  * Loads the config file and returns a settings struct
  */
```
avoidr, line 62: Remove this comment, it's already in `settings.h` and it doesn't belong here, anyway.

tests/mock/mock_settings.c
```diff
@@ -60,32 +60,13 @@ SETTINGS settings = {
...
-UTOX_SAVE *config_load(void) {
+void config_load(void) {
     FAIL_FATAL("called a mocked function, this should not happen: %s", __FUNCTION__);
 }
 
 /*
  * Writes save_in to the disk
  */
```
avoidr, line 69: Remove comment, same as above.

src/settings.c
```diff
@@ -283,327 +288,248 @@ static int config_parser(const char* section, const char* key, const char* value
...
 
+    char *last_slash = strrchr(config_path, '/');
+    if (!last_slash) {
+        last_slash = strrchr(config_path, '\\');
+    }
+
+    char *save_folder = strdup(config_path);
+    save_folder[last_slash - config_path + 1] = '\0';
+
+    if (!native_create_dir((uint8_t *)save_folder)) {
```
avoidr, line 326: This function is declared in `filesys.h` and defined in
```
android/main.c
posix/filesys.c
windows/filesys.c
```

`utox_get_filepath` wrongly creates the dir path on POSIX, but not on Windows.

I think this block (the commit) is fine, but put it in a separate function.

avoidr:
Generally, your commit messages don't have some info that you put in this PR. Priority should be on commit messages.

For example `Refactoring save and load uTox settings` is huge and could be split into some smaller pieces. But I don't think we're going to do that. We're going to go with amending commit message to at least mention that config is now written before Tox thread kills. And some other thing I saw.

`Remove unused function` → `Remove unused function config_osdefaults()`

src/main.c
redmanmale: `debug_file` is set via cmd args in `parse_args` which is called before `utox_init`.

avoidr: Yeah, right, I see now, thanks.

avoidr:
Wrote the macros here: d8ad7e9b2fbdf33ee96f72fee0b88e25f778fbb1

src/settings.c
```diff
@@ -380,9 +380,12 @@ static bool utox_save_config(void) {
...
     // notifications
-    WRITE_CONFIG_VALUE_BOOL(NOTIFICATIONS_SECTION, config->audible_notifications_enabled);
-    WRITE_CONFIG_VALUE_BOOL(NOTIFICATIONS_SECTION, config->status_notifications);
-    WRITE_CONFIG_VALUE_BOOL(NOTIFICATIONS_SECTION, config->no_typing_notifications);
+    WRITE_CONFIG_VALUE_BOOL(NOTIFICATIONS_SECTION,
+        config->audible_notifications_enabled);
+    WRITE_CONFIG_VALUE_BOOL(NOTIFICATIONS_SECTION,
+        config->status_notifications);
+    WRITE_CONFIG_VALUE_BOOL(NOTIFICATIONS_SECTION,
+        config->no_typing_notifications);
```
avoidr, line 388: I've considered that, but decided against it, by the way. It breaks reading flow, don't you think? The eye "trips" on that. I'd make this case an exception, or what do you think?

avoidr: I'd much rather change those variable names. `*_enabled` is unnecessary, for example. And `no_*` is inverted logic, not sure if great. Same goes for `disableudp`. But I wasn't sure if I wanted to fuss about that. Maybe we should.(?)

avoidr: And also, some variables have words delimited by `_`, some don't…

redmanmale: >I'd make this case an exception

Agreed.


Per names I also agree, but they are in users' config exactly like this.
We couldn't rename them just like that, it breaks loading values.

We have to create new params, add proxy logic (with old consts) to load old ones, deprecate them, and remove in ~0.19?.
It's a long way.

avoidr: Yes, I'm aware it's not a quick change. But let's do it. In a separate PR.

redmanmale: #1471.

avoidr:
I've split c1e1a5b63229ca018a388eabe73052f964650d0b (Refactoring save and load uTox settings) one last time on my branch: [avoidr/uTox:save-file-refactoring](https://github.com/avoidr/uTox/tree/save-file-refactoring)

src/settings.c
```diff
@@ -283,327 +297,219 @@ static int config_parser(const char* section, const char* key, const char* value
...
-    save->utox_last_version             = settings.curr_version;
-    save->group_notifications           = settings.group_notifications;
-    save->status_notifications          = settings.status_notifications;
-
-    save->language                      = settings.language;
-
-    memcpy(save->proxy_ip, proxy_address, proxy_address_size);
+void config_save(void) {
+    settings.save_version = UTOX_SAVE_VERSION;
+    settings.filter = flist_get_filter() == 1;
```
avoidr, line 506: `!!flist_get_filter()`?

avoidr:
I think it would also be good if in `Refactoring save and load uTox settings` you mentioned that you bumped `UTOX_SAVE_VERSION`, and that `UTOX_SAVE` was dropped. Or maybe not. I don't know anymore. It's not a requirement.

src/settings.c
redmanmale: I thing it's better to keep that way. Explicit is better than implicit and all that.

avoidr:
> > -    save->force_proxy                   = settings.force_proxy;
> -
> -    save->audio_device_in               = dropdown_audio_in.selected;
> -    save->audio_device_out              = dropdown_audio_out.selected;
> -    save->theme                         = settings.theme;
> -
> -    save->utox_last_version             = settings.curr_version;
> -    save->group_notifications           = settings.group_notifications;
> -    save->status_notifications          = settings.status_notifications;
> -
> -    save->language                      = settings.language;
> -
> -    memcpy(save->proxy_ip, proxy_address, proxy_address_size);
> +void config_save(void) {
> +    settings.save_version = UTOX_SAVE_VERSION;
> +    settings.filter = flist_get_filter() == 1;
> 
> I thing it's better to keep that way. Explicit is better than implicit and all that.

But `!!` is safer. `flist_get_filter` returns an int, and you only test for one positive value, leaving all other positive values to become 0. So the closest correct thing should be `!= 0`, but that's awkward. `!!` does the right thing, regardless of what type the function returns.

redmanmale:
I'll make this change.

But personally I don't like treating bool like int. Or vice-versa.
It has only two possible values: 0 and 1. I'd better make it bool.

But we should see in the future where we could have more values...
I think since this is our code and not some library API we should keep it as simple as it could be.
Two values? Bool.
Third is on the way? No problem, we'll make it int or some enum.

avoidr:
> But personally I don't like treating bool like int. Or vice-versa.
> It has only two possible values: 0 and 1. I'd better make it bool.

Yes, I think I agree. It's also more specific, so it's auto-documenting. Both, for humans, and compilers.

> But we should see in the future where we could have more values...
> I think since this is our code and not some library API we should keep it as simple as it could be.
> Two values? Bool.
> Third is on the way? No problem, we'll make it int or some enum.

Yes, yes.

I think after this `!!` we're finally done here. Let's make this bool change separately.

Btw, if `flist_get_filter` should really only return 0 or 1, then maybe it can be renamed to something that sounds more like bool… or maybe it should be something entirely else, I don't know, haven't looked at this function recently.

avoidr:
I rebased and reworded to add a newline after "Refactoring save and load uTox settings".

PR #1476, opened 2020-06-20
Update REAMDE and INSTALL

redmanmale:
* clean up a bit, drop obsolete links
* move Windows section up, fix table of contents

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1476)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,15 +8,12 @@ You shouldn't have to choose a new client for every device: So µTox works every
 
 ## Downloads
 - **Windows**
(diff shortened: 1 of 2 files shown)


avoidr:
> +Also [Jenkins](https://build.tox.chat) offers automatically compiled binaries.

Are the Jenkins builds still a thing? They are way old. Do we want to point there?

> +Installing on windows isn't really a thing yet... just download and run the latest version: [x64](https://github.com/uTox/uTox/releases/download/v0.17.2/utox_x86_64.exe).

s/windows/Windows/

> +Install uTox from the [Google Play]

s/the //

PR #1480, opened 2020-06-26
Settings fixes

avoidr:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1480)

### DIFF
--- a/src/settings.h
+++ b/src/settings.h
@@ -76,7 +76,7 @@ typedef struct utox_settings {
     uint32_t window_baseline;
     bool     window_maximized;
 
-    uint8_t proxy_ip[];
+    uint8_t proxy_ip[255]; /* coincides with TOX_MAX_HOSTNAME_LENGTH from toxcore */
 } SETTINGS;
 
 extern SETTINGS settings;
--- a/src/settings.h
+++ b/src/settings.h
@@ -20,14 +20,6 @@ typedef struct utox_settings {
 
     bool show_splash;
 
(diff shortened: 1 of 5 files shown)


avoidr:
I'm good here. Just reworded a commit message.

src/settings.h
```diff
@@ -76,7 +68,14 @@ typedef struct utox_settings {
...
 
-    uint8_t proxy_ip[];
+    // Low level settings (network, profile, portable-mode)
+    bool     portable_mode;
+    bool     disableudp;
+    bool     enableipv6;
+    bool     proxyenable;
+    bool     force_proxy;
+    uint16_t proxy_port;
+    uint8_t proxy_ip[255]; /* coincides with TOX_MAX_HOSTNAME_LENGTH from toxcore */
```
redmanmale, line 78: Make it `char` and all casts will go away.
And vertical alignment.

avoidr:
> +    uint8_t proxy_ip[255]; /* coincides with TOX_MAX_HOSTNAME_LENGTH from toxcore */
> 
> Make it `char` and all casts will go away.

I was thinking about that, but `char` is misleading. The data that goes in there is not ASCII, it's potentially Unicode. `proxy_ip` is not just an IP address, it may be a hostname. `proxy_ip` needs to be renamed to `proxy_host`, like it is in toxcore, too.

> And vertical alignment.
Thanks!

avoidr:
Rebased.

avoidr:
Are you fine with this PR?

PR #1485, opened 2020-07-10
Spring Cleaning

avoidr:
- [ ] remove includes from headers
- [ ] clean up includes
- [ ] refactor to remove including from parent directory

This resolves cyclic dependencies and should speed up building. Includes will automatically document the .c file, by virtue of only including the headers they need.


GrayHatter:
ew

PR #1487, opened 2020-07-22
fix dragging friend list (#1470)

### LINKED ISSUE #1470: Crash when a friend list item is dragged outside of friend list
Tried to fix this with 3d763b73de0fa2aa380f87fda177d6c27f7a9254, but then the UI broke and I had to revert.

The UI breakage was that items in the friend list didn't get deselected when the mouse pointer left the list, and then when one clicked on a tab in the settings, the last selected (or "not deselected") item was opened instead.

I tried it this way too: 9bcfc8c960e3a78c42bf2e1c5e076593e1c7420f
But it felt wrong, so I decided to just revert and do nothing, since this issue doesn't really come up in normal usage, anyway.

w23rd:
You should show the page for the item that was under the pointer when you clicked, not the one that is when you are moving the pointer.

Fixes #1470

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1487)

### DIFF
--- a/src/flist.c
+++ b/src/flist.c
@@ -66,7 +66,7 @@ static ITEM *right_mouse_item;
 
 static bool mouse_in_list;
 static bool selected_item_mousedown;
-static bool selected_item_mousedown_move_pend;
+static ITEM *selected_item_mousedown_move_pend;
 
 static int selected_item_dy; // y offset of selected item being dragged from its original position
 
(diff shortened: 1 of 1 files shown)


src/flist.c
```diff
@@ -1315,8 +1315,8 @@ bool flist_mup(void *UNUSED(n)) {
     bool draw = false;
     tooltip_mup(); /* may need to return one true */
 
-    if (mouseover_item && selected_item_mousedown_move_pend == true) {
-        show_page(mouseover_item);
+    if (mouseover_item && selected_item_mousedown_move_pend != NULL) {
```
avoidr, line 1318: Does the `mouseover_item` still make sense here?

w23rd: It depends on the behavior we expect. I supposed that if the pointer is not on an item, then the user moved the item unintentionally.

avoidr:
Thank you very much!

PR #1492, opened 2020-11-24
groups: load group topics from Tox save

avoidr:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1492)

### DIFF
--- a/src/groups.c
+++ b/src/groups.c
@@ -53,18 +53,18 @@ static GROUPCHAT *group_make(uint32_t group_number) {
     return &group[group_number];
 }
 
-GROUPCHAT *group_create(uint32_t group_number, bool av_group) {
+GROUPCHAT *group_create(uint32_t group_number, bool av_group, const char *name) {
     GROUPCHAT *g = group_make(group_number);
     if (!g) {
         LOG_ERR("Groupchats", "Could not get/create group %u", group_number);
         return NULL;
     }
 
-    group_init(g, group_number, av_group);
+    group_init(g, group_number, av_group, name);
     return g;
 }
 
-void group_init(GROUPCHAT *g, uint32_t group_number, bool av_group) {
+void group_init(GROUPCHAT *g, uint32_t group_number, bool av_group, const char *name) {
     pthread_mutex_lock(&messages_lock); /* make sure that messages has posted before we continue */
     if (!g->peer) {
         g->peer = calloc(UTOX_MAX_GROUP_PEERS, sizeof(GROUP_PEER *));
@@ -74,8 +74,13 @@ void group_init(GROUPCHAT *g, uint32_t group_number, bool av_group) {
         }
     }
 
-    snprintf((char *)g->name, sizeof(g->name), "Groupchat #%u", group_number);
-    g->name_length = strnlen(g->name, sizeof(g->name) - 1);
+    if (!name) {
+        snprintf(g->name, sizeof(g->name), "Groupchat #%u", group_number);
+        g->name_length = strlen(g->name);
+    } else {
+        snprintf(g->name, sizeof(g->name), "%s", name);
+        g->name_length = strlen(g->name);
+    }
 
     g->topic_length = sizeof("Drag friends to invite them") - 1;
     memcpy(g->topic, "Drag friends to invite them", sizeof("Drag friends to invite them") - 1);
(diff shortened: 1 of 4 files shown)


src/groups.c
```diff
@@ -53,18 +53,18 @@ static GROUPCHAT *group_make(uint32_t group_number) {
     return &group[group_number];
 }
 
-GROUPCHAT *group_create(uint32_t group_number, bool av_group) {
+GROUPCHAT *group_create(uint32_t group_number, bool av_group, char *name) {
```
robinlinden, line 56: const char *?

src/groups.c
```diff
@@ -53,18 +53,18 @@ static GROUPCHAT *group_make(uint32_t group_number) {
...
         return NULL;
     }
 
-    group_init(g, group_number, av_group);
+    group_init(g, group_number, av_group, name);
     return g;
 }
 
-void group_init(GROUPCHAT *g, uint32_t group_number, bool av_group) {
+void group_init(GROUPCHAT *g, uint32_t group_number, bool av_group, char *name) {
```
robinlinden, line 67: const char *?

src/groups.c
```diff
@@ -74,8 +74,13 @@ void group_init(GROUPCHAT *g, uint32_t group_number, bool av_group) {
         }
     }
 
-    snprintf((char *)g->name, sizeof(g->name), "Groupchat #%u", group_number);
-    g->name_length = strnlen(g->name, sizeof(g->name) - 1);
+    if (!name) {
+        snprintf((char *)g->name, sizeof(g->name), "Groupchat #%u", group_number);
```
robinlinden, line 78: Useless cast

src/groups.c
```diff
@@ -74,8 +74,13 @@ void group_init(GROUPCHAT *g, uint32_t group_number, bool av_group) {
...
+        g->name_length = strnlen(g->name, sizeof(g->name) - 1);
```
robinlinden, line 79: strnlen isn't in the C standard. You can (probably) use strnlen_s if uTox targets C11. (Or use strlen since you know g->name always is null-terminated.)

src/groups.c
```diff
@@ -74,8 +74,13 @@ void group_init(GROUPCHAT *g, uint32_t group_number, bool av_group) {
...
+    } else {
+        snprintf(g->name, sizeof(g->name), "%s", name);
+        g->name_length = strnlen(g->name, sizeof(g->name) - 1);
```
robinlinden, line 82: strnlen isn't in the C standard. You can (probably) use strnlen_s if uTox targets C11. (Or use strlen since you know g->name always is null-terminated.)

src/groups.c
```diff
@@ -317,6 +322,18 @@ void raze_groups(void) {
     group = NULL;
 }
 
+static size_t group_get_title_size(Tox *tox, uint32_t group_number) {
```
robinlinden, line 325: const Tox *?

src/groups.c
```diff
@@ -317,6 +322,18 @@ void raze_groups(void) {
...
+    size_t title_size;
```
robinlinden, line 326: This isn't C89. Declare and initialize on the same line.

robinlinden (changes requested):

src/groups.c
avoidr: Yes.

src/groups.c
avoidr: Right, I should have.

src/groups.c
avoidr: Drats! You're right, it's only POSIX.1-2008!

Yes, personally I would have left it out, too. Thanks.

src/groups.c
avoidr: Yes.

src/groups.c
avoidr: C isn't Python, either. That's the main reason why all of uTox is messy and looks like shit and I get aggressive reading it.

avoidr: I don't know if I have the strength to do this.

avoidr:
By the way, would you const also ints? Or only pointers? And why?

avoidr:
Rebased.

robinlinden:
> 
> 
> By the way, would you const also ints? Or only pointers? And why?

I think it's more important to const pointers as changes to them aren't local to the function. If it's a non-const pointer, I'd expect the function to write to it.

avoidr:
> > By the way, would you const also ints? Or only pointers? And why?
> 
> I think it's more important to const pointers as changes to them aren't local to the function. If it's a non-const pointer, I'd expect the function to write to it.

Ok, right, that's what I thought. I had confused constant pointer with
pointer to constant value. Thanks.

siburec:
I have tried this branch and I have seen the following:
1. If I create a group with <develop> branch and I change the name, when I build the <group_load_topic> branch I can see the modified name.
2. If I create a new group with <group_load_topic> version, the next time I open utox, the correct name is loaded.
3. However, if create a group with <develop> branch, build <group_load_topic> version and then open utox and change the name, this name is not detected by utox and continues showing Groupchat #x in the next opening.

avoidr:
You mean `3.` might be a bug? What exactly do you do? Step by step. And what do you expect?

avoidr:
I think I understand. I will take a look at it later. Thanks.

PR #1498, opened 2020-12-07
Fix group notifications

avoidr:


src/flist.c
```diff
@@ -108,17 +108,17 @@ static void flist_draw_name(ITEM *i, int x, int y, int width, char *name, char *
     }
 }
 
-static void flist_draw_status_icon(uint8_t status, int x, int y, bool notify) {
+static void flist_draw_status_icon(uint8_t status, int x, int y, bool notify, bool has_mention) {
```
GrayHatter, line 111: shouldn't `notify` be an enum instead of adding a 2nd bool? @avoidr

avoidr: Right, it probably should be, and it wouldn't even conflict that mentions don't make sense for friends.
Thanks.

avoidr: I added a commit, but I don't actually like that enum. It makes things overcomplicated and harder to understand. Now I have an enum that I need to explain, but the bool is self-explanatory. The scope is minimal, too, the enum doesn't help anything.

PR #1504, opened 2020-12-22
Copy text from chat without nicks via CTRL-C in Windows

redmanmale:
the same as in Unix.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1504)

### DIFF
--- a/src/windows/events.c
+++ b/src/windows/events.c
@@ -256,7 +256,7 @@ LRESULT CALLBACK WindowProc(HWND window, UINT msg, WPARAM wParam, LPARAM lParam)
             }
 
             if (control && wParam == 'C') {
-                copy(1);
+                copy(0);
                 return false;
             }
 


avoidr:
Does Windows have "copy with names" in its context menu?

avoidr:
You can actually say in commit message that it's now the same as in Unix.

redmanmale:
>Does Windows have "copy with names" in its context menu?

Yes, but it's very annoying.

avoidr:
What is annoying? I just want to know if users still have a way to copy with names after this patch.

Also, I think `s/username/nicks/` could be better. Shorter, and clearer.
And maybe also `s/shortcut/ctrl-c/`?

redmanmale:
I mean "yes, there's a context menu item to copy with names".
This patch is to use "copy without names" with ctrl-c.

avoidr:
My point is, can users still copy with names after this patch? It seems so, via right-click. The sane behaviour is to not copy names via ctrl-c, which this patch does. So this patch is fine and you're free to merge (preferably without merge commit for single commits).

PR #1511, opened 2021-01-07
Create a compilation database by default

krobelus:
This is useful for Clang-based tools, in particular language servers like
clangd. I think it's a good idea to provide it by default to improve the
development experience.  I didn't find a good way to detect a development
setup, so I just made it the default and disable it in scripts that are
mostly for CI. Sadly this turned out a bit crazy :(

If CMake read environment variables like Make, then a developer could use
"export CMAKE_EXPORT_COMPILE_COMMANDS=ON" (once, for all projects).
Is there a proper way to do something like that in CMake?

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1511)


robinlinden:
Funny you mention CMake reading env vars like that! CMAKE_EXPORT_COMPILE_COMMANDS is actually initialized from the env var CMAKE_EXPORT_COMPILE_COMMANDS. :D (In CMake 3.17 and newer)

krobelus:
Argh, I thought I had tried `CMAKE_EXPORT_COMPILE_COMMANDS=ON cmake .` (without consulting the docs) but I must have made a mistake.. so it actually works as expected! :open_mouth: 

I'm not really happy with the patch because it's so noisy.
It looks like the default behavior in most projects is to not generate compile_commands.json, so I guess the safe choice would be to reject this change. Then again, it doesn't hurt.

avoidr:
I will close this, since what this changes is already possible, but thanks for telling me about clangd, I didn't know it before :P

Thanks!

PR #1512, opened 2021-01-07
Disable -Werror for 3rd party projects

krobelus:
third_party/stb compiles with (scary) warnings. We should fix that but it
should not prevent a -Werror build.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1512)


avoidr:
Thanks!

PR #1513, opened 2021-01-07
Pass enum value with correct type when cancelling file transfers

krobelus:
Found by warnings like

src/file_transfers.c: In function ‘incoming_file_callback_request’:
src/file_transfers.c:881:59: error: implicit conversion from ‘enum <anonymous>’ to ‘TOX_FILE_CONTROL’ [-Werror=enum-conversion]
  881 |         tox_file_control(tox, friend_number, file_number, TOX_FILE_CANCEL, NULL);
      |                                                           ^~~~~~~~~~~~~~~

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1513)


avoidr:
This one actually fixes a real bug, thanks!

PR #1514, opened 2021-01-07
Use correct type for pthread entry point

krobelus:
This way we don't need to cast. Seems reasonable given that we want to treat
all warnings as errors. I got this one:

tests/mock/mock_threads.c:8:41: error: cast between incompatible function types from ‘void (*)(void *)’ to ‘void * (*)(void *)’ [-Werror=cast-function-type]
    8 |     pthread_create(&thread_temp, &attr, (void *(*)(void *))func, args);

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1514)

### DIFF
--- a/src/tox.c
+++ b/src/tox.c
@@ -452,7 +452,7 @@ static int init_toxcore(Tox **tox) {
  *
  * Accepts and returns nothing.
  */
-void toxcore_thread(void *UNUSED(args)) {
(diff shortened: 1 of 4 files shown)


zoff99:
thanks for the PR, i recently did this on my local build

avoidr:
Thanks!

Issue #1072, opened 2017-09-29
Crash on input after ctrl-z


redmanmale:
**Steps to reproduce**
* type some text into any edit
* cancel some text input via CTRL-Z
* type again

If that's not working try to type and cancel couple times.

**Env**
Win 7, 8.1, uTox v.0.16.1 and latest dev

fcore117:
[utox_p_debug_log.txt](https://github.com/uTox/uTox/files/1346411/utox_p_debug_log.txt)

redmanmale:
Crashed [here](https://github.com/uTox/uTox/blob/develop/src/ui/edit.c#L372) but it seems that memory is already corrupted there.

g4jc:
ASAN:
```
==7839==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602002df47b8 at pc 0x555555658d93 bp 0x7fffffffd060 sp 0x7fffffffd050
READ of size 8 at 0x602002df47b8 thread T0
    #0 0x555555658d92 in edit_do (/usr/bin/utox+0x104d92)
    #1 0x555555663713 in edit_char (/usr/bin/utox+0x10f713)
    #2 0x5555556b042a  (/usr/bin/utox+0x15c42a)
    #3 0x5555555d59e3 in main (/usr/bin/utox+0x819e3)
    #4 0x7ffff707c8e9 in __libc_start_main (/lib64/libc.so.6+0x218e9)
    #5 0x5555555d62e9 in _start (/usr/bin/utox+0x822e9)

0x602002df47b8 is located 0 bytes to the right of 8-byte region [0x602002df47b0,0x602002df47b8)
allocated by thread T0 here:
    #0 0x7ffff72b8580 in __interceptor_realloc /var/tmp/portage/sys-devel/gcc-7.3.0-r6/work/gcc-7.3.0/libsanitizer/asan/asan_malloc_linux.cc:75
    #1 0x555555658ace in edit_do (/usr/bin/utox+0x104ace)
    #2 0x555555663713 in edit_char (/usr/bin/utox+0x10f713)
    #3 0x5555556b042a  (/usr/bin/utox+0x15c42a)
    #4 0x5555555d59e3 in main (/usr/bin/utox+0x819e3)
    #5 0x7ffff707c8e9 in __libc_start_main (/lib64/libc.so.6+0x218e9)

SUMMARY: AddressSanitizer: heap-buffer-overflow (/usr/bin/utox+0x104d92) in edit_do
[...]

PR #1523, opened 2021-01-19
Solved crash when undoing chat input with ctrl-z

siburec:
Solved a bug that caused a crash when text from chat was undone using ctrl-z.
It should solve issue #1072.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1523)

### DIFF
--- a/src/ui/edit.c
+++ b/src/ui/edit.c
@@ -354,13 +354,17 @@ static uint16_t edit_change_do(EDIT *edit, EDIT_CHANGE *c) {
 void edit_do(EDIT *edit, uint16_t start, uint16_t length, bool remove) {
     EDIT_CHANGE *new, **history;
 
+    if (edit->history_cur != edit->history_length) {
+        uint16_t i = edit->history_cur;
+        while (i != edit->history_length) {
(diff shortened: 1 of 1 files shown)


siburec:
If somebody wants to reproduce the error, it happens when there are 2 o more characters written in a chat and ctrl-z is pressed twice and then a new character is written.
It happened because, when a new character is added, edit_do() function is called and it frees the memory of the removed characters. However it is done after reallocating only edit->history_cur + 1 values, so after 2 ctrl-z edit->history_cur = edit->history_length -2 and therefore last value hadn't been reallocated.
For that reason, the memory should be freed before reallocating it.

avoidr:
Thanks! I will make time in the morning to review.

>It happened because, when a new character is added, edit_do() function is called and it frees the memory of the removed characters. However it is done after reallocating only edit->history_cur + 1 values, so after 2 ctrl-z edit->history_cur = edit->history_length -2 and therefore last value hadn't been reallocated.
>For that reason, the memory should be freed before reallocating it.

Another perfect candidate for the commit message :P

avoidr:
Thanks!

PR #1525, opened 2021-01-24
Fix double free in flist when changing settings w/ empty flist

robinlinden:
This pretty much only happens if you double-click e.g. the udp toggle to
trigger a restart of tox twice in rapid succession. The first restart
calls flist_freeall() which empties the flist and leaves old pointers in
`item` and `shown_list` and the second flist_freeall() will try to free
the old pointers again.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1525)

### DIFF
--- a/src/flist.c
+++ b/src/flist.c
@@ -820,6 +820,11 @@ void flist_delete_rmouse_item(void) {
 }
 
 void flist_freeall(void) {
+    if (!item) {
+        LOG_ERR("F-List", "Friend list is already NULL!");
+        return;
+    }
+
     for (ITEM *i = item; i != item + itemcount; i++) {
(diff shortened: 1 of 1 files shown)


avoidr:
Maybe we should skip the for-loop altogether if `item == NULL`? We're only missing the NULL-deref by the fact that `itemcount == 0`, but that feels very implicit to me. Actually, I think an early return would be the best.

avoidr:
Otherwise perfect. I've been clicking settings furiously for minutes without a single crash!

GrayHatter:
:LGTM:

avoidr:
Thanks!

PR #1527, opened 2021-01-28
Fix memleaks

avoidr:
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1527)

### DIFF
--- a/src/friend.c
+++ b/src/friend.c
@@ -500,6 +500,7 @@ void friend_add(char *name, uint16_t length, char *msg, uint16_t msg_length) {
         memcpy(data, id, TOX_PUBLIC_KEY_SIZE);
         postmessage_toxcore(TOX_FRIEND_NEW_NO_REQ, TOX_PUBLIC_KEY_SIZE, 0, data);
         addfriend_status = ADDF_NOFREQUESTSENT;
+        free(data);
     } else {
         addfriend_status = ADDF_BADNAME;
     }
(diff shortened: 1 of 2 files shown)


src/friend.c
```diff
@@ -500,6 +500,7 @@ void friend_add(char *name, uint16_t length, char *msg, uint16_t msg_length) {
         memcpy(data, id, TOX_PUBLIC_KEY_SIZE);
         postmessage_toxcore(TOX_FRIEND_NEW_NO_REQ, TOX_PUBLIC_KEY_SIZE, 0, data);
         addfriend_status = ADDF_NOFREQUESTSENT;
+        free(data);
```
robinlinden, line 503: This isn't right. Ownership is transferred to the Tox thread when you call postmessage_toxcore. This will probably lead to a crash due to a double free.

src/groups.c
```diff
@@ -359,6 +359,8 @@ void init_groups(Tox *tox) {
         free(title);
     }
     LOG_INFO("Groupchat", "Initialzied groupchat array with %u groups", self.groups_list_size);
+
+    free(group);
```
robinlinden, line 363: This also isn't right. This is a static variable that's used in other functions (e.g. `group_make` and `group_create`) and freed only in `raze_groups`.

robinlinden (changes requested):

avoidr:
Right, that makes sense. I reacted to "calloc without free". Lesson learned. Thank you for your review.

PR #1540, opened 2021-07-21
IRC: Move to libera.chat away from freenode

avoidr:
Freenode has become distrustworthy and the channel is no more under our
control.

---
This change is [Reviewable](https://reviewable.io/reviews/utox/utox/1540)


GrayHatter:
LGTM

PR #1570, opened 2024-11-25
fix some bugs on Mac:

yaozhi20:
1. crash when voice
2. startup gui not freshed

### DIFF
--- a/src/cocoa/main.m
+++ b/src/cocoa/main.m
@@ -363,8 +363,8 @@ void launch_at_startup(bool should) {
                             self.utox_window.contentView =
                                 [[[uToxView alloc] initWithFrame:(CGRect){ 0, 0, self.utox_window.frame.size }] autorelease];
 
-                            ui_size(settings.window_width, settings.window_height);
                             ui_rescale(ui_scale);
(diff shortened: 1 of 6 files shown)


src/cocoa/main.m
```diff
@@ -495,7 +495,7 @@ int main(int argc, char const *argv[]) {
     setlocale(LC_ALL, "");
 
     /* set the width/height of the drawing region */
-    ui_size(settings.window_width, settings.window_height);
+    // ui_size(settings.window_width, settings.window_height);
```
GrayHatter, line 498: why did this get moved?

GrayHatter (changes requested):
As I can't test this, because I can't build on mac, can you provide a few more details, about where you found these bugs, and how you tested the fixes?

src/cocoa/main.m
yaozhi20: this is not correct ,now I find out. 
But other files modified  is needed . I found these when try to build and use uTox on my Mac.
ANd i repull a commit (https://github.com/uTox/uTox/pull/1576). Thank your reply.

Issue #1574, opened 2024-12-05
Crash on launch with `XLib tray: Reached end of function, this is bad juju!`


barracuda156:
### Overview

1. uTox version: 0.18.1
2. Operating system: macOS 10.6
3. Desktop Environment/Window Manager (Linux/BSD only): XQuartz 1.20.10

### Issue

I am trying to fix X11 build on macOS. (Video disabled due to relying on `vid4l`.)
I can get the build completing, though at the moment I need to do linking manually, since CMake errs out for some reason, and the app launches with GUI. However, it is unusable: a few seconds into the GUI, and it crashes with:
```
36-174% utox -v
XLib Tray:    Incoming tray window event (28)
XLib tray:    Reached end of function, this is bad juju!
terminate called after throwing an instance of 'std::system_error'
  what():  Resource temporarily unavailable
```

Any idea what may be the issue?

barracuda156:
Looks like it crashes on audio:
```
Process:         utox-orig [2112]
Path:            /opt/local/bin/utox-orig
Identifier:      utox-orig
Version:         ??? (???)
Code Type:       PPC (Native)
Parent Process:  zsh [2110]

Date/Time:       2024-12-06 01:56:50.100 +0800
OS Version:      Mac OS X 10.6 (10A190)
Report Version:  6

Exception Type:  EXC_BAD_ACCESS (SIGBUS)
Exception Codes: KERN_PROTECTION_FAILURE at 0x0000000000000024
Crashed Thread:  3

Thread 0:
0   libSystem.B.dylib             	0x007ff974 select$DARWIN_EXTSN + 12
1   libxcb.1.dylib                	0x0222fe94 _xcb_conn_wait + 272
2   libxcb.1.dylib                	0x02231684 xcb_wait_for_event + 80
3   libX11.6.dylib                	0x00568568 _XReadEvents + 188
4   libX11.6.dylib                	0x005573c8 XNextEvent + 76
5   utox-orig                     	0x00087f38 main + 2192
6   utox-orig                     	0x00005754 start + 64

Thread 1:
0   libSystem.B.dylib             	0x007a2e0c __semwait_signal + 8
1   libSystem.B.dylib             	0x007a2c20 nanosleep$UNIX2003 + 184
2   utox-orig                     	0x00018764 toxcore_thread + 2588

Thread 2:
0   libSystem.B.dylib             	0x007a2e0c __semwait_signal + 8
[...]

PR #1575, opened 2024-12-05
Allow X11 build on macOS

barracuda156:
UPDATE: the app confirmed to build and launch without crashes on 10.6.8 at least. I leave a passage below, but it is no longer relevant.

Opening as a draft, since while it builds, the app crashes on launch (GUI starts, and in a few seconds crashes), apparently due to OpenAL-related stuff. See: https://github.com/uTox/uTox/issues/1574
This happens regardless of whether Apple framework is used or `openal-soft`. Switching to either is easy, but neither helps.

It was a pain to sort out embedding the icon, since Linux way is non-portable. Since macOS needs a section name to be within 16 characters, I changed the file name to `icons/utox.png` (locally I just cloned `utox-128x128.png` to `utox.png`). The file is identical, but obviously it must be actually present. So if this ever gets merged, the icon file should either be renamed or a copy of it added.

Video is stubbed out, since it depends on `vid4l`, which is also non-portable, AFAIU.

I open this in a hope someone could join in to improve this, so that it actually works :)

### DIFF
--- a/src/xlib/video.c
+++ b/src/xlib/video.c
@@ -211,6 +211,7 @@ uint16_t native_video_detect(void) {
 static uint16_t video_x, video_y;
 
 bool native_video_init(void *handle) {
+#ifndef __APPLE__
     if (isdesktop(handle)) {
         utox_v4l_fd = -1;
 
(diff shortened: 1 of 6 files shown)


barracuda156:
This works on newer macOS, though does not on PowerPC.
So it can be merged, but I would like to have some feedback first.

barracuda156:
I have it working on PowerPC now:

utox

src/xlib/video.c
```diff
@@ -211,6 +211,7 @@ uint16_t native_video_detect(void) {
 static uint16_t video_x, video_y;
 
 bool native_video_init(void *handle) {
+#ifndef __APPLE__
```
GrayHatter, line 214: normally I'd prefer creating a new native entry, but I'm ok with this otherwise.

GrayHatter (approved):

GrayHatter:
I've asked a friend to double check my work, here and once they confirm I didn't miss anything stupid, I'll merge this within the day :)

barracuda156:
@GrayHatter Thank you for merging!

barracuda156:
@GrayHatter I just noticed my note about the icon file. Could you make a copy of the icon with a shortened name as described in the head topic (the code now assumes it exists)? If that is not acceptable, then I should add a note in the README, so that a user building from source makes a local copy with a command in terminal.

GrayHatter:
you can add a note to the readme, or have the build script make a copy of the icon

